PR #56, opened 2025-11-02
feat: add support for fluss tables  using  zstd compression

binary-signal:
Add ipc_compression feature to arrow dependency


binary-signal:
plus implement deserialize  for timestamps

PR #95, opened 2025-12-14
chore: upgrade opendal and adopt jiff

tisonkun:
Newer OpenDAL version uses jiff which is proper maintained with better documented features.

... also remove now unnecessary `once_cell` dep since they are provided from `std::sync`.


luoyuxia (approved):
@tisonkun Thank you very much. LGTM

PR #96, opened 2025-12-15
chore: use log over tracing

tisonkun:
Ref -

* https://fast.github.io/blog/fastrace-a-modern-approach-to-distributed-tracing-in-rust/
* https://www.tisonkun.org/2025/08/20/logforth/

And anyway tracing can be set to understand `log`'s output.

### DIFF
--- a/crates/fluss/Cargo.toml
+++ b/crates/fluss/Cargo.toml
@@ -37,24 +37,23 @@ futures = "0.3"
 clap = { workspace = true }
 crc32c = "0.6.8"
 linked-hash-map = "0.5.6"
-prost = "0.13.5"
+prost = "0.14"
 rand = "0.9.1"
 serde = { version = "1.0.219", features = ["derive", "rc"] }
 serde_json = "1.0.140"
-thiserror = "1.0"
-tracing = "0.1"
+thiserror = "2"
+log = { version = "0.4", features = ["kv_std"] }
 tokio = { workspace = true }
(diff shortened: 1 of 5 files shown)


crates/fluss/Cargo.toml
```diff
@@ -37,18 +37,18 @@ futures = "0.3"
...
-tracing = "0.1"
+thiserror = "2"
+log = { version = "0.4" }
 tokio = { workspace = true }
 parking_lot = "0.12"
 bytes = "1.10.1"
 dashmap = "6.1.0"
 rust_decimal = "1"
-ordered-float = { version = "4", features = ["serde"] }
+ordered-float = { version = "5", features = ["serde"] }
```
tisonkun: Hash format changed. But I think we're yet to rely on it.

crates/fluss/src/io/file_io.rs
```diff
@@ -97,12 +97,10 @@ impl FileIOBuilder {
     }
 }
 
-#[async_trait::async_trait]
 pub trait FileRead: Send + Unpin + 'static {
-    async fn read(&self, range: Range<u64>) -> Result<Bytes>;
+    fn read(&self, range: Range<u64>) -> impl Future<Output = Result<Bytes>> + Send;
 }
```
tisonkun: See https://blog.rust-lang.org/2023/12/21/async-fn-rpit-in-traits/#async-fn-in-public-traits

Issue #42, opened 2025-10-20
scanner should handle when bucket leader doesn't exist.


luoyuxia:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/fluss-rust/issues) and found nothing similar.

### Description

see https://github.com/apache/fluss-rust/blob/1b265353f7329cbc83b50e443aa44288899d370e/crates/fluss/src/client/table/scanner.rs#L243

### Willingness to contribute

- [ ] I'm willing to submit a PR!

leekeiabstraction:
@luoyuxia 

Do we aim to just achieve the same behaviour as [java's implementation of LogFetcher.java where fetch requests are skipped if leader is not available](https://github.com/apache/fluss/blob/main/fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/LogFetcher.java#L488-L491)?

Happy to contribute if you can clarify.

PR #102, opened 2025-12-20
[Scanner] Handle leader availability

leekeiabstraction:
### Purpose

Linked issue: close https://github.com/apache/fluss-rust/issues/42

### Brief change log

Add server unavailability handling in log scanner. Allows client to recover when leader fails over in various scenarios. 

### Tests

Manual tests performed on latest changes are as follow

|Scenario|Expected Behaviour|Testing Result|
|--|--|--|
|Start scanner with leader unavailable|Scanner connects to leader when leader becomes available| Passed|
|Single replica table leader becomes unavailable|Scanner connects to leader when leader becomes available| Passed|
|Multi replica table leader becomes unavailable|Scanner connects to new leader|Passed|

### DIFF
--- a/crates/fluss/src/rpc/server_connection.rs
+++ b/crates/fluss/src/rpc/server_connection.rs
@@ -72,7 +72,7 @@ impl RpcClient {
         };
 
         if let Some(conn) = connection {
-            if !conn.is_poisoned().await {
+            if !conn.is_poisoned() {
                 return Ok(conn);
             }
         }


luoyuxia:
> > @leekeiabstraction Thanks for the pr. Besides,
> > I think we should add https://github.com/apache/fluss/blob/main/fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/LogFetcher.java#L173
> > to update metadata if leader is not found. Othwise, we'll never be able to fetch the bucket if the the leader is not found at first. Retry should get the leader and make it's possible to fetch the bucket
> 
> I've pushed the changes to handle this specific case. However, I am not certain on how to simulate the particular scenario. When using a local cluster:
> 
> 1. Scenario A, stop tablet server in between polls: If I understand correctly, the current Java implementation of `checkAndUpdateMetadata()` (of which we're following) does not update metadata if bucket leader was previously available, it only updates it if bucket leader was previously unavailable. Additionally, [LogFetcher.send_fetches() currently just logs fetch failures](https://github.com/apache/fluss-rust/blob/main/crates/fluss/src/client/table/scanner.rs#L372-L376), leader status is not updated.
> 2. Scenario B, start LogScanner with tablet server down: I run into issue before instantiating LogScanner when  calling admin.get_table(): `Error: FlussAPIError { api_error: ApiError { code: 17, message: "Replication factor: 1 larger than available tablet servers: 0." } }`.

1: yes, you're correct. It's not easy to mock leader not avaiable. You have to mock create lots of replica to enfore leader election cost a while. If hard, I think we can just ignore testing.

2: it looks werid, the exception should only be thrown while create table. I think there must be some issue in the fluss-rust client.

leekeiabstraction:
@luoyuxia 

Would appreciate review on the latest changes. I've tested with changing leader, all tablet servers down and then recovered scenario, the latest changes seem to be working fine with one exception:

We currently do not treat `ConnectionError("IO Error: Connection refused (os error 61)")` as poisoned, I think we should. In my manual testing (at least on MacOS), once a connection gets into this state, the connection is not recoverable even if the server behind it is running again.

leekeiabstraction:
Manual tests performed on latest changes are as follow

|Scenario|Expected Behaviour|Testing Result|
|--|--|--|
|Start scanner with leader unavailable|Scanner connects to leader when leader becomes available| Passed|
|Single replica table leader becomes unavailable|Scanner connects to leader when leader becomes available| Passed|
|Multi replica table leader becomes unavailable|Scanner connects to new leader|Passed|

@luoyuxia Would appreciate a review here. Thank you!

PR #116, opened 2025-12-26
chore: add integration tests for subscribe_batch and project_by_name

AndreaBozzo:
- Add `test_subscribe_batch`: tests `LogScanner::subscribe_batch()` with HashMap
- Add `test_project_by_name`: tests `TableScan::project_by_name()` with column names
- Both tests include error case validation

Linked issue: #31

### DIFF
--- a/crates/fluss/tests/integration/table.rs
+++ b/crates/fluss/tests/integration/table.rs
@@ -36,13 +36,16 @@ mod table_test {
     use crate::integration::fluss_cluster::{FlussTestingCluster, FlussTestingClusterBuilder};
     use crate::integration::utils::create_table;
     use arrow::array::record_batch;
+    use fluss::client::{FlussTable, TableScan};
     use fluss::metadata::{DataTypes, Schema, TableBucket, TableDescriptor, TablePath};
+    use fluss::record::ScanRecord;
     use fluss::row::InternalRow;
     use fluss::rpc::message::OffsetSpec;
     use jiff::Timestamp;
     use std::collections::HashMap;
     use std::sync::Arc;
     use std::thread;
+    use std::time::Duration;
 
     fn before_all() {
         // Create a new tokio runtime in a separate thread
@@ -138,6 +141,11 @@ mod table_test {
 
         append_writer.flush().await.expect("Failed to flush");
 
+        // Create scanner to verify appended records
+        let table = connection
+            .get_table(&table_path)
+            .await
+            .expect("Failed to get table");
         let num_buckets = table.table_info().get_num_buckets();
         let log_scanner = table
             .new_scan()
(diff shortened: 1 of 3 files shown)


crates/fluss/tests/integration/table.rs
```diff
@@ -446,22 +446,32 @@ mod table_test {
             .await
             .expect("Failed to poll");
 
-        let records: Vec<_> = scan_records.into_iter().collect();
+        let mut records: Vec<_> = scan_records.into_iter().collect();
+        records.sort_by_key(|r| r.offset());
```
leekeiabstraction: Curious, is this sort_by_key needed? What happens if you do not sort? 

My assumption is that poll_batch should already return records ordered by offset.

leekeiabstraction:
> The `sort_by_key` is needed because `poll()` does not guarantee ordering as records may arrive from different buckets or be processed asynchronously. Also thank you again on this becouse that's how i noticed the pattern!
> 
> This follows the existing pattern in the codebase:
> 
> * [table.rs:159](https://github.com/apache/fluss-rust/blob/main/crates/fluss/tests/integration/table.rs#L159) - `append_record_batch_and_scan`
> * [table.rs:189](https://github.com/apache/fluss-rust/blob/main/crates/fluss/tests/integration/table.rs#L189), same test with projection

You're right. Java side uses inAnyOrder assertion: https://github.com/apache/fluss/blob/bcebe7f48dde96f7085bf14af41ff18afc1f99e6/fluss-client/src/test/java/org/apache/fluss/client/table/scanner/log/RemoteLogScannerITCase.java#L114

Approving!

AndreaBozzo:
> > The `sort_by_key` is needed because `poll()` does not guarantee ordering as records may arrive from different buckets or be processed asynchronously. Also thank you again on this becouse that's how i noticed the pattern!
> > This follows the existing pattern in the codebase:
> > 
> > * [table.rs:159](https://github.com/apache/fluss-rust/blob/main/crates/fluss/tests/integration/table.rs#L159) - `append_record_batch_and_scan`
> > * [table.rs:189](https://github.com/apache/fluss-rust/blob/main/crates/fluss/tests/integration/table.rs#L189), same test with projection
> 
> You're right. Java side uses inAnyOrder assertion: https://github.com/apache/fluss/blob/bcebe7f48dde96f7085bf14af41ff18afc1f99e6/fluss-client/src/test/java/org/apache/fluss/client/table/scanner/log/RemoteLogScannerITCase.java#L114
> 
> Approving!

Kudos Mr.Lee !

AndreaBozzo:
> @AndreaBozzo Thanks for the pr. LGTM! I only append a small commit to improve test.

Thanks @luoyuxia , looking forward to further help on this promising project, take care !

luoyuxia:
> > @AndreaBozzo Thanks for the pr. LGTM! I only append a small commit to improve test.
> 
> Thanks @luoyuxia , looking forward to further help on this promising project, take care !

Cool! Look forward to it.

Issue #109, opened 2025-12-22
Support ZSTD compression level


leekeiabstraction:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/fluss-rust/issues) and found nothing similar.

### Motivation

We found that column pruning breaks because compression was not implemented on rust client.

This was fixed by adding compression support (with default ZSTD compression level), now we should add the missing compression level support: https://github.com/apache/fluss-rust/pull/108#discussion_r2639093134

### Willingness to contribute

- [x] I'm willing to submit a PR!

PR #120, opened 2025-12-29
Chore: Rust client to return error when currently unimplemented non-default ZSTD compression is configured

leekeiabstraction:
### Purpose

Related: https://github.com/apache/fluss-rust/issues/109

### Brief change log

Fail fast by returning error when currently unimplemented non-default ZSTD compression is configured

### Tests

Added UTs


luoyuxia (approved):
@leekeiabstraction LGTM. I just append a small commit to refine the error

PR #138, opened 2026-01-10
chore: implement zero copy in CompactedRowReader

luoyuxia:
### Purpose

Linked issue: close #xxx

### DIFF
--- a/crates/fluss/src/row/compacted/compacted_row.rs
+++ b/crates/fluss/src/row/compacted/compacted_row.rs
@@ -26,11 +26,11 @@ use crate::row::{GenericRow, InternalRow};
 #[allow(dead_code)]
 pub struct CompactedRow<'a> {
     arity: usize,
-    segment: &'a [u8],
     size_in_bytes: usize,
     decoded_row: OnceLock<GenericRow<'a>>,
     deserializer: CompactedRowDeserializer<'a>,
     reader: CompactedRowReader<'a>,
+    data_types: &'a [DataType],
 }
 
 pub fn calculate_bit_set_width_in_bytes(arity: usize) -> usize {
@@ -39,16 +39,16 @@ pub fn calculate_bit_set_width_in_bytes(arity: usize) -> usize {
 
 #[allow(dead_code)]
 impl<'a> CompactedRow<'a> {
-    pub fn from_bytes(types: &'a [DataType], data: &'a [u8]) -> Self {
-        let arity = types.len();
+    pub fn from_bytes(data_types: &'a [DataType], data: &'a [u8]) -> Self {
+        let arity = data_types.len();
         let size = data.len();
         Self {
             arity,
-            segment: data,
             size_in_bytes: size,
             decoded_row: OnceLock::new(),
-            deserializer: CompactedRowDeserializer::new(types),
+            deserializer: CompactedRowDeserializer::new(data_types),
             reader: CompactedRowReader::new(arity, data, 0, size),
+            data_types,
         }
     }
 
@@ -69,7 +69,7 @@ impl<'a> InternalRow for CompactedRow<'a> {
     }
 
     fn is_null_at(&self, pos: usize) -> bool {
-        self.decoded_row().is_null_at(pos)
+        self.data_types[pos].is_nullable() && self.reader.is_null_at(pos)
     }
 
     fn get_boolean(&self, pos: usize) -> bool {
--- a/crates/fluss/src/row/compacted/compacted_row_reader.rs
+++ b/crates/fluss/src/row/compacted/compacted_row_reader.rs
@@ -204,8 +204,8 @@ impl<'a> CompactedRowReader<'a> {
         (val, next_pos)
     }
 
-    pub fn read_binary(&self, length: usize) -> (&'a [u8], usize) {
-        self.read_bytes(length)
+    pub fn read_binary(&self, pos: usize) -> (&'a [u8], usize) {
+        self.read_bytes(pos)
     }
 
     pub fn read_bytes(&self, pos: usize) -> (&'a [u8], usize) {


crates/fluss/src/row/datum.rs
```diff
@@ -296,9 +292,7 @@ impl Datum<'_> {
             Datum::Float32(v) => append_value_to_arrow!(Float32Builder, v.into_inner()),
             Datum::Float64(v) => append_value_to_arrow!(Float64Builder, v.into_inner()),
             Datum::String(v) => append_value_to_arrow!(StringBuilder, *v),
-            Datum::OwnedString(v) => append_value_to_arrow!(StringBuilder, v.as_str()),
-            Datum::Blob(v) => append_value_to_arrow!(BinaryBuilder, v.as_ref()),
-            Datum::BorrowedBlob(v) => append_value_to_arrow!(BinaryBuilder, *v),
```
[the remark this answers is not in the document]
luoyuxia: Let me first merge your pr since `Owned` also accept reference and value which is more flexible and I can build based on your pr.

luoyuxia:
> I’m not convinced the transmute approach is sound here. CompactedRow<'a> is public and 'a is unconstrained (constructors don’t tie it to any input), so a caller can choose 'a = 'static. With mem::transmute we can then manufacture 'static references into row.segment, and after the row is dropped those references dangle -> use-after-free / UB. This isn’t about address stability, but about lifetime validity.
> 
> I strongly prefer the Cow approach for Datum (Cow, Cow<[u8]>): it keeps the borrow checker honest, avoids unsafe, and still allows zero-copy via Borrowed when the backing buffer is known to outlive the use-site, while preserving Owned for Python/async boundaries.

Thanks for the input. Agreed, let me have a refactor again

luoyuxia:
@Kelvinyu1117 @fresh-borzoni @leekeiabstraction Comments has been addressed. PTAL

fresh-borzoni (commented):
Nice direction 👍 

I'm worried about this: Cell<usize> makes the reader mutate itself during decoding. If decoding stops half-way, the reader stays advanced and later reads can be wrong. 
Better: keep the cursor (pos) local to deserialize() and keep the reader immutable.

crates/fluss/src/row/compacted/compacted_row_reader.rs
```diff
@@ -167,55 +156,58 @@ impl CompactedRowReader {
         panic!("Invalid input stream.");
```
leekeiabstraction: Same q here

leekeiabstraction: Please ignore. Shouldn't review on my bed 🤦🏻‍♂️

crates/fluss/src/row/compacted/compacted_row_reader.rs
```diff
@@ -71,151 +92,130 @@ impl CompactedRowDeserializer {
...
-        debug_assert!(self.position + 2 <= self.limit);
-        let bytes_slice = &self.segment[self.position..self.position + 2];
-        let byte_array: [u8; 2] = bytes_slice
-            .try_into()
-            .expect("Slice must be exactly 2 bytes long");
-
-        self.position += 2;
-        i16::from_ne_bytes(byte_array)
+    pub fn read_short(&self, pos: usize) -> (i16, usize) {
+        debug_assert!(pos + 2 <= self.limit);
```
leekeiabstraction: We can calculate pos + x once and assign to variable. Similarly on other methods

leekeiabstraction (commented):
Thank you for the commits, just a couple last comments.

crates/fluss/src/row/compacted/compacted_row_reader.rs
```diff
@@ -151,7 +140,7 @@ impl CompactedRowReader {
         panic!("Invalid input stream.");
```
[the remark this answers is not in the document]
luoyuxia: The most signication bit of the i32 must be zero

crates/fluss/src/row/compacted/compacted_row_reader.rs
```diff
@@ -15,54 +15,75 @@
...
 
-    pub fn deserialize(&self, reader: &mut CompactedRowReader) -> GenericRow<'static> {
+    pub fn deserialize(&self, reader: &CompactedRowReader<'a>) -> GenericRow<'a> {
         let mut row = GenericRow::new();
-        for (pos, dtype) in self.schema.iter().enumerate() {
-            if reader.is_null_at(pos) {
-                row.set_field(pos, Datum::Null);
+        let mut cursor = reader.initial_position();
+        for (col_pos, dtype) in self.schema.iter().enumerate() {
+            if reader.is_null_at(col_pos) {
```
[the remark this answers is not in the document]
luoyuxia: Let's just check the data type is nullable just like java side.

crates/fluss/src/row/compacted/compacted_row_reader.rs
```diff
@@ -71,151 +92,133 @@ impl CompactedRowDeserializer {
...
-    pub fn read_binary(&mut self, length: usize) -> Bytes {
-        debug_assert!(self.position + length <= self.limit);
-
-        let start = self.position;
-        let end = start + length;
-        self.position = end;
-
-        self.segment.slice(start..end)
+    pub fn read_binary(&self, length: usize) -> (&'a [u8], usize) {
+        self.read_bytes(length)
```
fresh-borzoni: Why do we path length when position is expected?

crates/fluss/src/row/compacted/compacted_row.rs
```diff
@@ -24,125 +24,95 @@ use crate::row::{GenericRow, InternalRow};
 // Reference implementation:
 // https://github.com/apache/fluss/blob/main/fluss-common/src/main/java/org/apache/fluss/row/compacted/CompactedRow.java
 #[allow(dead_code)]
-pub struct CompactedRow {
+pub struct CompactedRow<'a> {
     arity: usize,
-    segment: Bytes,
-    offset: usize,
+    segment: &'a [u8],
```
fresh-borzoni: Do we still use it?

crates/fluss/src/row/compacted/compacted_row.rs
```diff
@@ -24,125 +24,95 @@ use crate::row::{GenericRow, InternalRow};
...
-    fn decoded_row(&mut self) -> &GenericRow<'static> {
-        if !self.decoded {
-            self.reader
-                .point_to(self.segment.clone(), self.offset, self.size_in_bytes);
-            self.decoded_row = self.deserializer.deserialize(&mut self.reader);
-            self.decoded = true;
-        }
-        &self.decoded_row
+    fn is_null_at(&self, pos: usize) -> bool {
+        self.decoded_row().is_null_at(pos)
```
fresh-borzoni: We shouldn't pay deserialization price for just checking for null bitmap.

The only problem is semantical one, as we use this in deserialize:
```rust
if dtype.is_nullable() && reader.is_null_at(col_pos)
```

fresh-borzoni: maybe we want to store types in CompactedRow as well?
then we may use it like this:
```rust
 self.types[pos].is_nullable() && self.reader.is_null_at(pos)
 ```

luoyuxia:
@fresh-borzoni Thanks for review. Comments is addressed.

PR #139, opened 2026-01-10
Improve datum, use Cow to reduce verbosity and avoid Box::leak gymnastics

leekeiabstraction:
### Purpose

- Improve datum, use Cow to reduce verbosity and avoid necessity of Box::leak gymnastics


luoyuxia (commented):
@leekeiabstraction Thank for thr pr!

PR #147, opened 2026-01-11
[ISSUE-146] decrement flushes_in_progress counter in await_flush_comp…

### LINKED ISSUE #146: flushes_in_progress counter never decrements, breaking batching after first flush in Accumulator
### Search before asking

### Please describe the bug 🐞

The `flushes_in_progress` counter is incremented in `begin_flush()` but never decremented in `await_flush_completion()`. This causes batching to be permanently disabled after the first flush.

  ## Root Cause
  This is a porting bug from the Java implementation. The Java code correctly decrements in a `finally` block:

  **Java (correct):**
  ```java
  public void awaitFlushCompletion() throws InterruptedException {
      try {
          for (WriteBatch.RequestFuture future : incomplete.requestResults()) {
              future.await();
          }
      } finally {
          flushesInProgress.decrementAndGet();  // ← Cleanup
      }
  }
```

  Rust (incorrect):
```rust
  pub async fn await_flush_completion(&self) -> Result<()> {
      let handles: Vec<_> = self.incomplete_batches.read().values().cloned().collect();
      for result_handle in handles {
          result_handle.wait().await?;
      }
      Ok(())
      // ← Missing decrement!
  }
```

  Impact

  After the first flush() call:
  - flushes_in_progress = 1 forever
  - flush_in_progress() returns true forever
  - In batch_ready(): all batches become immediately sendable
  - Batching optimization is permanently disabled

  References

  - Fluss source: https://github.com/apache/fluss/blob/main/fluss-client/src/main/java/org/apache/fluss/client/write/RecordAccumulator.java#L362-L374
  - Kafka source: https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/clients/producer/internals/RecordAccumulator.java#L1097-L1113

### Solution

Add decrement after await loop completes (in all code paths).

### Are you willing to submit a PR?

fresh-borzoni:
## Summary
  Fixes critical bug where `flushes_in_progress` counter is never decremented, causing batching to be permanently disabled after the first flush.

  Closes #146

  ## The Bug
  `begin_flush()` increments the counter but `await_flush_completion()` never decrements it. This is a porting error from the Java implementation which correctly decrements in a `finally` block.

  ## Impact
  After the first `flush()` call:
  - `flushes_in_progress` stays at 1 forever
  - `flush_in_progress()` always returns `true`
  - All batches become immediately sendable (no batching optimization)

 ## The Fix
 Always decrement counter after flush completes (mimics Java `finally` block)

### DIFF
--- a/crates/fluss/src/client/write/accumulator.rs
+++ b/crates/fluss/src/client/write/accumulator.rs
@@ -31,6 +31,9 @@ use std::sync::Arc;
 use std::sync::atomic::{AtomicI32, AtomicI64, Ordering};
 use tokio::sync::Mutex;
 
+// Type alias to simplify complex nested types
+type BucketBatches = Vec<(BucketId, Arc<Mutex<VecDeque<WriteBatch>>>)>;
+
 #[allow(dead_code)]
 pub struct RecordAccumulator {
     config: Config,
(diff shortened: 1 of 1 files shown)


fresh-borzoni:
I'll rebase once  another branch is merged.

leekeiabstraction (commented):
Well spotted!

Can we add an assertion on some of our integration test to check that flush_in_progress() returns false?

fresh-borzoni:
@leekeiabstraction Thanks for the review.

I added a unit test that proves the fix is sound. Integration test would require making the method public and threading it through, which I don't feel is needed at this point.

leekeiabstraction:
LGTM thank you!

luoyuxia (approved):
@fresh-borzoni Thanks for that. LGTM! Forget the decrement while implement fluss side.

fresh-borzoni:
@luoyuxia No worries, my pleasure to help

merged main here, good to merge now

Issue #113, opened 2025-12-22
Support upsert in FlussTable


luoyuxia:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/fluss-rust/issues) and found nothing similar.

### Description

refer to java side https://github.com/apache/fluss/blob/de2d9a042637386f9c8f59bddbff8a1a66853ca0/fluss-client/src/main/java/org/apache/fluss/client/table/Table.java#L74 to make put kv work end to end

### Willingness to contribute

- [ ] I'm willing to submit a PR!

leekeiabstraction:
@luoyuxia Please can you assign this to me?

PR #158, opened 2026-01-13
Task: WriteFormat and various small changes to unblock Upsert implementation

leekeiabstraction:
Related: #113

### DIFF
--- a/crates/fluss/src/client/write/write_format.rs
+++ b/crates/fluss/src/client/write/write_format.rs
@@ -22,15 +22,13 @@ use std::fmt::Display;
 
 pub enum WriteFormat {
     ArrowLog,
-    IndexedLog,
     CompactedLog,
-    IndexedKv,
     CompactedKv,
 }
 
 impl WriteFormat {
     pub const fn is_log(&self) -> bool {
-        matches!(self, Self::ArrowLog | Self::IndexedLog | Self::CompactedLog)
+        matches!(self, Self::ArrowLog | Self::CompactedLog)
     }
 
     pub fn is_kv(&self) -> bool {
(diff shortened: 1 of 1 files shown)


leekeiabstraction:
@luoyuxia @fresh-borzoni @Kelvinyu1117 

Would appreciate review here. I'm breaking up TableUpsert implementation PR into smaller bits so that it is more easily reviewable. 

Will follow up with CompactedRowEncoder PR before UpsertTable PR.

crates/fluss/src/metadata/table.rs
```diff
@@ -726,6 +728,22 @@ impl TableConfig {
...
     }
+
+    pub fn get_datalake_format(&self) -> Result<Option<DataLakeFormat>> {
+        self.properties
+            .get("table.datalake.format")
+            .map(|f| f.parse().map_err(Error::from))
+            .transpose()
+    }
+
+    pub fn get_kv_format(&self) -> Result<Option<KvFormat>> {
```
fresh-borzoni: Why do we return Optional if we have default and unwrap?
Should it match get_datalake_format logic or not?

crates/fluss/src/client/write/write_format.rs
```diff
@@ -0,0 +1,68 @@
...
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+use crate::error::Error::IllegalArgument;
+use crate::error::Result;
+use crate::metadata::KvFormat;
+use std::fmt::Display;
+
+pub enum WriteFormat {
```
luoyuxia: nit:
only include `ArrowLog`, `CompactedLog`, `CompactedKv`.
Indexedxx is deperated and will be drop in the future

luoyuxia (commented):
@leekeiabstraction Thanks. Left minor comments.

PR #159, opened 2026-01-14
[client] Add lookup support for primary key tables

### LINKED ISSUE #114: Support lookup
### Search before asking

### Description

Lookup is needed to verify put kv. successfull or not.

### Willingness to contribute

AndreaBozzo:
### Purpose

Linked issue: close #114

Implements lookup functionality to verify put kv operations. This follows the simple lookup approach as suggested in the issue comments (without complex batching/queuing).

### Brief change log

- Add `Lookup` API key (1017) to `api_key.rs` - matching [Java Fluss ApiKeys.java](https://github.com/apache/fluss/blob/main/fluss-rpc/src/main/java/org/apache/fluss/rpc/protocol/ApiKeys.java#L48)
- Add lookup proto messages (`LookupRequest`, `LookupResponse`, `PbLookupReqForBucket`, `PbLookupRespForBucket`, `PbValue`) to `fluss_api.proto`
- Create `lookup.rs` message module with `LookupRequest` struct implementing `RequestBody` trait
- Add async `lookup(bucket_id, key)` method to `FlussTable` that returns `Option<Vec<u8>>`

Reference Java implementation:
- [LookupSender.java](https://github.com/apache/fluss/blob/main/fluss-client/src/main/java/org/apache/fluss/client/lookup/LookupSender.java)
- [PrimaryKeyLookuper.java](https://github.com/apache/fluss/blob/main/fluss-client/src/main/java/org/apache/fluss/client/lookup/PrimaryKeyLookuper.java)
- [TabletService.java lookup()](https://github.com/apache/fluss/blob/main/fluss-server/src/main/java/org/apache/fluss/server/tablet/TabletService.java#L236-L253)

### API and Format

Adds new public API:
- `FlussTable::lookup(bucket_id: i32, key: Vec<u8>) -> Result<Option<Vec<u8>>>`

### Documentation

No documentation changes required at this time.

### DIFF
--- a/crates/fluss/src/client/table/lookup.rs
+++ b/crates/fluss/src/client/table/lookup.rs
@@ -177,7 +177,8 @@ impl<'a> Lookuper<'a> {
     /// # Returns
     /// * `Ok(LookupResult)` - The lookup result (may be empty if key not found)
     /// * `Err(Error)` - If the lookup fails
-    pub async fn lookup(&mut self, row: &dyn InternalRow) -> Result<LookupResult> {
+    pub async fn lookup(&mut self, row: &dyn InternalRow) -> Result<LookupResult<'_>> {
+        // todo: support batch lookup
         // Encode the key from the row
         let encoded_key = self.key_encoder.encode_key(row)?;
         let key_bytes = encoded_key.to_vec();
--- a/crates/fluss/src/record/kv/kv_record_batch_builder.rs
+++ b/crates/fluss/src/record/kv/kv_record_batch_builder.rs
@@ -485,7 +485,6 @@ mod tests {
 
     #[test]
     fn test_builder_with_compacted_row_writer() {
-        use crate::metadata::{DataType, IntType, StringType};
         use crate::record::kv::KvRecordBatch;
         use crate::row::InternalRow;
         use crate::row::compacted::CompactedRow;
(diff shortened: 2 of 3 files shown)


AndreaBozzo:
@luoyuxia good morning, i didn't know if u were already working on this, i'm sorry if its the case

@leekeiabstraction mr.lee as you are proficient in java and i'm not it would be really appreciated if you can check this out too

Have a nice day, open to feedbacks/changes

luoyuxia (commented):
@AndreaBozzo Thanks for the pr. Just one comment to keep api align with java side.

crates/fluss/src/client/table/lookup.rs
```diff
@@ -0,0 +1,170 @@
...
+    /// bucketing function.
+    ///
+    /// # Arguments
+    /// * `key` - The encoded primary key bytes
+    ///
+    /// # Returns
+    /// * `Ok(Some(Vec<u8>))` - The value bytes if the key exists
+    /// * `Ok(None)` - If the key does not exist
+    /// * `Err(Error)` - If the lookup fails
+    pub async fn lookup(&self, key: Vec<u8>) -> Result<Option<Vec<u8>>> {
```
luoyuxia: use  `row: &dyn InternalRow` instead of `key: Vec<u8>`, we should do the encoding for users.

Also, we will need `LookupResult` just like java intead of `Option<Vec<u8>>` to handle lookups mutiple rows. 

LookupResult should contains `InternalRow` instead of `Vec<u8>`, a byte array is hard to use for user. But for simplity, you can just use `CompactedRow`

AndreaBozzo:
Rebased and addressed last comments from luo, leaving also edits on ON so he can have maximum control over this since it grew quite a bit from my initial implementation.

Thanks for the help!

leekeiabstraction (approved):
LGTM, nit (haven't left comment) that we should leave TODO comments in places where we know implementation is not at parity w/ Java side.

Approving, as this PR will unblock testing for my later PR as well.

PR #189, opened 2026-01-21
chore: invalidate leader info on stale metadata error when handling fetch response

zhaohaidao:
### Purpose

Linked issue: close #xxx

### DIFF
--- a/crates/fluss/src/client/table/scanner.rs
+++ b/crates/fluss/src/client/table/scanner.rs
@@ -36,7 +36,7 @@ use crate::client::table::remote_log::{
     RemoteLogDownloader, RemoteLogFetchInfo, RemotePendingFetch,
 };
 use crate::error::{ApiError, Error, FlussError, Result};
-use crate::metadata::{TableBucket, TableInfo, TablePath};
+use crate::metadata::{PhysicalTablePath, TableBucket, TableInfo, TablePath};
 use crate::proto::{ErrorResponse, FetchLogRequest, PbFetchLogReqForBucket, PbFetchLogReqForTable};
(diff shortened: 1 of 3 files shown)


zhaohaidao:
@luoyuxia Hi, PTAL if u have time.

crates/fluss/src/cluster/cluster.rs
```diff
@@ -101,6 +101,31 @@ impl Cluster {
...
+        let mut available_locations_by_bucket = self.available_locations_by_bucket.clone();
+        available_locations_by_bucket.remove(table_bucket);
+
+        let mut available_locations_by_path = self.available_locations_by_path.clone();
+        if let Some(table_path) = self.table_path_by_id.get(&table_bucket.table_id()) {
+            if let Some(locations) = available_locations_by_path.get_mut(table_path) {
+                locations.retain(|location| location.table_bucket() != table_bucket);
+            }
+        } else {
+            for locations in available_locations_by_path.values_mut() {
```
fresh-borzoni: It's defensive and good, but if `table_id` isn't in `table_path_by_id`, that indicates a data structure inconsistency and we hide some issues with this fallback.
Do we wish to add a warn here at least?

crates/fluss/src/client/table/scanner.rs
```diff
@@ -745,6 +758,9 @@ impl LogFetcher {
                     .into();
 
                     let error = FlussError::for_code(error_code);
+                    if Self::should_invalidate_bucket_leader(error) {
```
leekeiabstraction: Should we consider having the check and invalidation in other places e.g. sender or lookup?

PR #220, opened 2026-01-28
Implement partitioning in Upsert / Lookup

### LINKED ISSUE #178: Implement partitioning in Upsert / Lookup
### Search before asking

### Description

Partitioning in Upsert / Lookup is currently left as unimplemented/todo. We should implement it to achieve parity with Java side

### Willingness to contribute

leekeiabstraction:
### Purpose

Linked issue: close #178 

### Brief change log

- Add partitioning support in Upsert / Lookup
- Add Integration test
- Added example

### DIFF
--- a/crates/fluss/src/client/table/lookup.rs
+++ b/crates/fluss/src/client/table/lookup.rs
@@ -249,7 +249,7 @@ impl<'a> Lookuper<'a> {
         let cluster = self.metadata.get_cluster();
         let leader = self
             .metadata
-            .leader_for(self.table_info.get_table_path(), &table_bucket)
+            .leader_for(self.table_path.as_ref(), &table_bucket)
             .await?
             .ok_or_else(|| Error::LeaderNotAvailable {
                 message: format!("No leader found for table bucket: {table_bucket}"),
--- a/crates/fluss/src/client/write/accumulator.rs
+++ b/crates/fluss/src/client/write/accumulator.rs
@@ -195,17 +195,11 @@ impl RecordAccumulator {
 
     pub async fn ready(&self, cluster: &Arc<Cluster>) -> Result<ReadyCheckResult> {
         // Snapshot just the Arcs we need, avoiding cloning the entire BucketAndWriteBatches struct
-        let entries: Vec<(
-            Arc<PhysicalTablePath>,
-            bool,
-            Option<PartitionId>,
-            BucketBatches,
-        )> = self
+        let entries: Vec<(Arc<PhysicalTablePath>, Option<PartitionId>, BucketBatches)> = self
             .write_batches
             .iter()
             .map(|entry| {
                 let physical_table_path = Arc::clone(entry.key());
-                let is_partitioned_table = entry.value().is_partitioned_table;
                 let partition_id = entry.value().partition_id;
                 let bucket_batches: Vec<_> = entry
                     .value()
@@ -213,12 +207,7 @@ impl RecordAccumulator {
                     .iter()
                     .map(|(bucket_id, batch_arc)| (*bucket_id, batch_arc.clone()))
                     .collect();
-                (
-                    physical_table_path,
-                    is_partitioned_table,
-                    partition_id,
-                    bucket_batches,
-                )
+                (physical_table_path, partition_id, bucket_batches)
             })
             .collect();
 
@@ -226,12 +215,11 @@ impl RecordAccumulator {
         let mut next_ready_check_delay_ms = self.batch_timeout_ms;
         let mut unknown_leader_tables = HashSet::new();
 
-        for (physical_table_path, is_partitioned_table, mut partition_id, bucket_batches) in entries
-        {
+        for (physical_table_path, mut partition_id, bucket_batches) in entries {
             next_ready_check_delay_ms = self
                 .bucket_ready(
                     &physical_table_path,
-                    is_partitioned_table,
+                    physical_table_path.get_partition_name().is_some(),
                     &mut partition_id,
                     bucket_batches,
                     &mut ready_nodes,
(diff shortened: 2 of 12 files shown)


crates/fluss/src/client/table/lookup.rs
```diff
@@ -187,26 +213,46 @@ impl<'a> Lookuper<'a> {
     /// * `Err(Error)` - If the lookup fails
     pub async fn lookup(&mut self, row: &dyn InternalRow) -> Result<LookupResult<'_>> {
         // todo: support batch lookup
-        // Encode the key from the row
-        let encoded_key = self.key_encoder.encode_key(row)?;
-        let key_bytes = encoded_key.to_vec();
+        let pk_bytes = self.primary_key_encoder.encode_key(row)?.to_vec();
+        let bk_bytes = match &mut self.bucket_key_encoder {
+            Some(encoder) => &encoder.encode_key(row)?.to_vec(),
```
fresh-borzoni: Wouldn't this temporary Vec be freed at the end of match, so we would return reference to something that would be freed?

crates/fluss/src/client/write/accumulator.rs
```diff
@@ -185,34 +190,73 @@ impl RecordAccumulator {
...
+                } else {
+                    log::debug!(
+                        "Partition does not exist for {}, bucket will not be set to ready",
+                        physical_table_path.as_ref()
+                    );
+
+                    // TODO: we shouldn't add unready partitions to unknownLeaderTables,
+                    // because it cases PartitionNotExistException later
+                    unknown_leader_tables.insert(physical_table_path);
+                    return Ok(ReadyCheckResult::new(
```
fresh-borzoni: why return? shouldn't we continue?

fresh-borzoni (commented):
@leekeiabstraction Thanks for the PR! Left comments. PTAL

crates/fluss/src/client/write/accumulator.rs
leekeiabstraction: This is similar to Java flow: https://github.com/apache/fluss/blob/96faf08cbd84d1d43a2ef610c189829fa9c34e76/fluss-client/src/main/java/org/apache/fluss/client/write/RecordAccumulator.java#L463-L469

Returning is fine because sender.run_once() resolves the unknown table leader and on subsequent run, goes through this flow again.

crates/fluss/src/client/table/lookup.rs
fresh-borzoni: Okay, borrow checker is smarter than I thought :) 
But still why do we copy to return reference. Mb just `as_ref()`?

leekeiabstraction: Ah, I see what you mean now.

crates/fluss/src/metadata/table.rs
```diff
@@ -373,7 +374,7 @@ impl TableDescriptorBuilder {
         self
     }
 
-    pub fn partitioned_by(mut self, partition_keys: Vec<String>) -> Self {
+    pub fn partitioned_by(mut self, partition_keys: Arc<[String]>) -> Self {
```
luoyuxia: why change this. From user side,  pass a `Vec<String>` should be simpler than pass `Arc`, right?

crates/fluss/src/client/write/sender.rs
```diff
@@ -343,11 +378,12 @@ impl Sender {
                         .error_message()
                         .cloned()
                         .unwrap_or_else(|| error.message().to_string());
-                    if let Some(table_path) = self
+                    if let Some((table_path, physical_table_path)) = self
```
luoyuxia: `physical_table_path` also contains `table_path`, so do the tunple is required? Can just return `physical_table_path`?

crates/fluss/src/metadata/table.rs
```diff
@@ -749,19 +750,19 @@ impl TablePath {
 /// `partition_name` will be `Some(...)`; otherwise, it will be `None`.
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
 pub struct PhysicalTablePath {
-    table_path: TablePath,
+    table_path: Arc<TablePath>,
```
luoyuxia: why make it as `Arc`?  I think `TablePath` should be good.

crates/fluss/src/metadata/table.rs
leekeiabstraction: I went overboard with refactoring here. Well spotted. Will address in follow up commit

leekeiabstraction:
All comments addressed. PTAL @luoyuxia

luoyuxia (commented):
@leekeiabstraction Thanks for the update. Only one more comments. Otherwise, LGTM

crates/fluss/src/client/write/sender.rs
```diff
@@ -392,15 +393,15 @@ impl Sender {
...
+                if let Some(physical_table_path) = self
                     .handle_write_batch_error(
                         ready_batch,
                         FlussError::UnknownServerError,
                         format!("Missing response for table bucket {bucket}"),
                     )
                     .await?
                 {
-                    invalid_metadata_tables.insert(table_path);
+                    invalid_metadata_tables.insert(physical_table_path.get_table_path().clone());
```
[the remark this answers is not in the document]
leekeiabstraction: Java side updates based on both TablePath and PhysicalTablePath: https://github.com/apache/fluss/blob/3a803dfd812398decbfbced81b655d6a8839351e/fluss-client/src/main/java/org/apache/fluss/client/metadata/MetadataUpdater.java#L231-L240

Here we have `invalid_metadata_tables` which tracks invalid TablePath and `invalid_physical_table_paths` which tracks invalid PhysicalTablePath

PR #258, opened 2026-02-06
[TASK-240] Fire-and-forget behaviour for efficient batching

### LINKED ISSUE #240: Rust UpsertWriter::upsert() and delete() block waiting for server acknowledgment on each call
### Search before asking

### Description

Remove `.wait()` from `upsert()`/`delete()`, let `flush()` handle ack waiting. This enables true concurrent batching.
Also it matches Java's fire-and-forget behaviour with CompletableFutures

### Willingness to contribute

fresh-borzoni:
## Summary
 
 closes  #240 
 
 Fire-and-forget write semantics for efficient batching                                                                                                                                                                                          
                                                                
  Write operations (`upsert`, `delete`, `append`) now return a `WriteResultFuture` immediately after queueing the record, instead of blocking until server acknowledgment. This enables efficient batching and records accumulate in the               
  `RecordAccumulator` and are sent in batches by the background `Sender`, similar to rdkafka's `DeliveryFuture`.
                                                                                                                                                                                                                                                     
  ### Usage patterns                                                               

  **Fire-and-forget** — queue writes, flush at end:
  ```rust
  writer.upsert(&row1).await?;
  writer.upsert(&row2).await?;
  writer.flush().await?; // ensures all writes are acknowledged
```

  Per-record acknowledgment — double-await for read-after-write consistency:
```rust
  writer.upsert(&row).await?.await?;
```

Notes:
  - Rust core: upsert(), delete(), append(), append_arrow_batch() return Result<WriteResultFuture> instead of Result<()>/Result<UpsertResult>
  - Python bindings: Methods return an awaitable ack handle via nested future_into_py, both UpsertWriter and AppendWriter support both patterns
  - C++ bindings: Added WriteResult class with Wait() method; AppendWriter::Append() has two overloads — fire-and-forget (single arg) and with ack handle (out parameter)
  - Examples and tests: Updated across all languages to demonstrate both patterns

### DIFF
--- a/bindings/cpp/examples/example.cpp
+++ b/bindings/cpp/examples/example.cpp
@@ -91,6 +91,7 @@ int main() {
         {3, "Charlie", 92.1f, 35},
     };
 
+    // Fire-and-forget: queue rows, flush at end
     for (const auto& r : rows) {
         fluss::GenericRow row;
         row.SetInt32(0, r.id);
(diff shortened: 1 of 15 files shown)


fresh-borzoni:
We may wish to refactor queueing to use sync to avoid `await?` on queueing, it would match rdkafka, but better in separate followup PR.

PR #264, opened 2026-02-06
chore: python bindings cleanup and bit of refactor

fresh-borzoni:
## Summary

Refactor LogScanner to use ScannerKind enum                                                                                                                                                                                                     
                                                                
  Replace two `Option<T>` fields (`inner`, `inner_batch`) on `LogScanner` with a                                                                                                                                                                     
  single `ScannerKind` enum. The previous design allowed impossible states, both                                                                                                                                                                    
  fields `None`, or both `Some` — which required defensive `if/else` chains in                                                                                                                                                                       
  every method. The enum makes the active scanner variant explicit at the type     
  level.

### DIFF
--- a/bindings/python/src/table.rs
+++ b/bindings/python/src/table.rs
@@ -265,33 +265,28 @@ impl TableScan {
             let (projected_schema, projected_row_type) =
                 calculate_projected_types(&table_info, projection_indices)?;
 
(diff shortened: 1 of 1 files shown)


fresh-borzoni:
@leekeiabstraction @luoyuxia PTAL 🙏

PR #294, opened 2026-02-09
[TASK-287] Support projection by names in CPP

### LINKED ISSUE #287: support project by name in cpp binding
### Search before asking

### Description

as title. Only project by index is supported.

### Willingness to contribute

fresh-borzoni:
closes #287 
                                                                                                                                                                                                                                                     
  - Add ProjectByName to CPP TableScan, bringing parity with Rust and Python

  // By index
  table.NewScan().ProjectByIndex{0, 7}).CreateLogScanner(scanner);

  // By name (new)
  table.NewScan().ProjectByName({"id", "updated_at"}).CreateLogScanner(scanner);

### DIFF
--- a/bindings/cpp/src/table.cpp
+++ b/bindings/cpp/src/table.cpp
@@ -129,7 +129,7 @@ TableScan Table::NewScan() { return TableScan(table_); }
 // TableScan implementation
 TableScan::TableScan(ffi::Table* table) noexcept : table_(table) {}
 
-TableScan& TableScan::Project(std::vector<size_t> column_indices) {
+TableScan& TableScan::ProjectByIndex(std::vector<size_t> column_indices) {
     projection_ = std::move(column_indices);
     name_projection_.clear();
     return *this;
--- a/bindings/cpp/examples/example.cpp
+++ b/bindings/cpp/examples/example.cpp
@@ -197,7 +197,7 @@ int main() {
     std::vector<size_t> projected_columns = {0, 7};
     fluss::LogScanner projected_scanner;
     check("new_log_scanner_with_projection",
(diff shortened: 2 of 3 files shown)


bindings/cpp/src/table.cpp
```diff
@@ -129,25 +129,54 @@ TableScan Table::NewScan() { return TableScan(table_); }
 // TableScan implementation
 TableScan::TableScan(ffi::Table* table) noexcept : table_(table) {}
 
-TableScan& TableScan::Project(std::vector<size_t> column_indices) {
+TableScan& TableScan::ProjectByIndex(std::vector<size_t> column_indices) {
```
leekeiabstraction: nit: Keep similar to Rust side with `Project`

bindings/cpp/src/table.cpp
```diff
@@ -163,15 +192,16 @@ Result TableScan::CreateRecordBatchScanner(LogScanner& out) {
     }
 
     try {
-        if (projection_.empty()) {
-            out.scanner_ = table_->new_record_batch_log_scanner();
-        } else {
+        auto resolved = !name_projection_.empty() ? ResolveNameProjection() : projection_;
```
leekeiabstraction: nit: resolved_indices

bindings/cpp/src/table.cpp
fresh-borzoni: Hmm, okay, though I think it's better to have symmetrical methods, because users usually think in names, so now it feels odd as if default is indices.

Should we change this in all clients to symmetrical methods?

luoyuxia: Thanks for the explanation. Make sense to me. Seems we should follow rust/python approach

PR #330, opened 2026-02-15
[TASK-87] Optimize copying in CPP bindings

### LINKED ISSUE #87: Optimize copying during C++ and Rust data conversion.
### Search before asking

### Description

Copying large blocks of data may cause performance bottlenecks. We should optimize the copying method if necessary.

### Willingness to contribute

fresh-borzoni:
## Summary

  closes #87

 Replace value-copying FFI layer with opaque Rust-backed types for zero-copy reads and cheaper writes in CPP bindings.                                                                                                                              
                  
  ## Changes                                                                                                                                                                                                                                         
                  
  - **Opaque types**: replace `Datum`/`FfiDatum`/`FfiGenericRow`/`FfiScanRecord` copying infrastructure with three opaque Rust types — `GenericRowInner` (writes), `ScanResultInner` (scan reads), `LookupResultInner` (lookup reads)
  - **Zero-copy reads**: new `RowView` class borrows directly from Rust scan results; `LookupResult` class replaces `(bool found, GenericRow out)` pattern
  
   ## Benchmark Results                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                     
  **Environment:** Apple M1 Max, 32 GB RAM, macOS 15.7.1, local fluss cluster                                                                                                                                                                                                                                                                                                                                                                                                
  **Workload:** 1M rows per run, averaged over 100 runs                                                                                                                                                                                              
                                                                                                                                                                                                                                                     
  Two scan configurations:                                                                                                                                                                                                                           
  - **8 mixed fields:** int, string, float, int, date, time, timestamp, timestamp_ltz
  - **4×100B strings:** int + 4 string fields (100 bytes each, ~381 MB raw data)

  Each scan polls all 1M records from the server, accumulates them in memory,
  then reads every field. Reported time covers both phases (poll + field access).

  Memory is measured as C++ heap delta (`malloc_zone_statistics` `size_in_use`)
  before and after accumulating all records.

  ### Throughput

  | Metric | main | opaque types | Change |
  |---|---|---|---|
  | Append fire-and-forget | 1,634 ms | 1,176 ms | **-28%** |
  | Append with ack | 1,662 ms | 1,109 ms | **-33%** |
  | Scan (8 mixed fields) | 639 ms | 222 ms | **-65%** |
  | Scan (4×100B strings) | 1,115 ms | 349 ms | **-69%** |

  ### Memory (held during scan)

  | Metric | main | opaque types | Change |
  |---|---|---|---|
  | Scan (8 mixed fields) | 915 MB | 119 MB | **-87%** |
  | Scan (4×100B strings) | 1,344 MB | 471 MB | **-65%** |

Memory figures reflect worst-case accumulation. In typical streaming usage memory stays flat, but per-batch overhead is still reduced since the intermediate FfiDatum copy layer is eliminated entirely.

### DIFF
--- a/bindings/cpp/src/table.cpp
+++ b/bindings/cpp/src/table.cpp
@@ -79,6 +79,12 @@ int Date::Day() const {
     return tm.tm_mday;
 }
 
+// NOLINTNEXTLINE(cppcoreguidelines-macro-usage)
+#define CHECK_INNER(name)                                                                 \
+    do {                                                                                  \
+        if (!inner_) throw std::logic_error(name ": not available (moved-from or null)"); \
+    } while (0)
+
 // ============================================================================
 // GenericRow — write-only row backed by opaque Rust GenericRowInner
 // ============================================================================
@@ -100,6 +106,7 @@ void GenericRow::Destroy() noexcept {
         rust::Box<ffi::GenericRowInner>::from_raw(inner_);
         inner_ = nullptr;
     }
+    column_map_.reset();
 }
 
 GenericRow::GenericRow(GenericRow&& other) noexcept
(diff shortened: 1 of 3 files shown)


fresh-borzoni:
@zhaohaidao @luoyuxia PTAL 🙏

bindings/cpp/src/lib.rs
```diff
@@ -1417,3 +1510,494 @@ impl LogScanner {
...
+            .get(field)
+            .map_or(0, |c| types::core_data_type_to_ffi(c.data_type()))
+    }
+
+    pub fn column_name(columns: &[fcore::metadata::Column], field: usize) -> &str {
+        columns.get(field).map_or("", |c| c.name())
+    }
+
+    pub fn is_null(row: &dyn InternalRow, _field: usize) -> bool {
+        row.is_null_at(_field)
```
leekeiabstraction: Rename `_field` to `field` as it is actually used

bindings/cpp/src/table.cpp
```diff
@@ -79,6 +79,321 @@ int Date::Day() const {
...
+    if (!column_map_) {
+        BuildColumnMap();
+    }
+    return column_map_;
+}
+
+ScanRecord ScanRecords::operator[](size_t idx) const {
+    if (!inner_ || idx >= inner_->sv_record_count())
+        return ScanRecord{0, std::nullopt,           0,
+                          0, ChangeType::AppendOnly, RowView(nullptr, 0, nullptr)};
```
leekeiabstraction: Should we consider applying fail-fast by throwing an exception here instead? Looks like operator[] is public and user can access it directly.

fresh-borzoni:
Addressed comments. 
Added simple benchmark table  to demonstrate the difference. It's with my local fluss cluster, so it's just for demonstration.

leekeiabstraction (commented):
Thank you for the PR! Not used to C++ so a bit out of water here but left some comments.

bindings/cpp/src/table.cpp
```diff
@@ -79,6 +79,321 @@ int Date::Day() const {
...
+// ============================================================================
+
+ScanRecords::ScanRecords() noexcept = default;
+
+ScanRecords::~ScanRecords() noexcept { Destroy(); }
+
+void ScanRecords::Destroy() noexcept {
+    if (inner_) {
+        rust::Box<ffi::ScanResultInner>::from_raw(inner_);
+        inner_ = nullptr;
```
zhaohaidao: Destroy() releases inner_ but does not clear column_map_, so reused objects keep stale mappings

bindings/cpp/include/fluss.hpp
```diff
@@ -799,48 +608,122 @@ struct GenericRow {
...
+};
 
-    void EnsureSize(size_t idx) {
-        if (fields.size() <= idx) {
-            fields.resize(idx + 1);
+/// Read-only row view for scan results. Zero-copy access to string and bytes data.
+///
+/// WARNING: RowView borrows from ScanRecords. It must not outlive the ScanRecords
+/// that produced it (similar to std::string_view borrowing from std::string).
+class RowView : public detail::NamedGetters<RowView> {
```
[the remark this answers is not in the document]
fresh-borzoni: Yes, I'll tweak docs after this code only PR.

zhaohaidao (commented):
@fresh-borzoni Thanks for your pr. left minor comments.

bindings/cpp/include/fluss.hpp
```diff
@@ -232,6 +238,14 @@ struct Timestamp {
...
 
+enum class ChangeType {
+    AppendOnly = 0,
+    Insert = 1,
+    UpdateBefore = 2,
+    UpdateAfter = 3,
+    Delete = 4,
+};
+
 enum class TypeId {
```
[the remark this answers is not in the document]
fresh-borzoni: Good idea in principle, but it would require including the CXX-generated header in fluss.hpp, which exposes the whole FFI layer and increases compilation time for consumers. 

It's nontrivial to work around, so I'd keep it out of this PR. Happy to explore it separately.

bindings/cpp/src/lib.rs
```diff
@@ -1632,128 +1632,247 @@ mod row_reader {
...
+        columns: &'a [fcore::metadata::Column],
+        field: usize,
+        getter: &str,
+        allowed: impl FnOnce(&fcore::metadata::DataType) -> bool,
+    ) -> Result<&'a fcore::metadata::DataType, String> {
+        let col = get_column(columns, field)?;
+        if row.is_null_at(field) {
+            return Err(format!("field {field} is null"));
+        }
+        let dt = col.data_type();
```
leekeiabstraction: Should this only become an error if data_type is nullable?

bindings/cpp/src/table.cpp
```diff
@@ -79,6 +79,369 @@ int Date::Day() const {
...
+}
+
+bool GenericRow::Available() const { return inner_ != nullptr; }
+
+void GenericRow::Reset() {
+    if (inner_) inner_->gr_reset();
+}
+
+void GenericRow::SetNull(size_t idx) {
+    if (inner_) inner_->gr_set_null(idx);
```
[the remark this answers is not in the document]
fresh-borzoni: Good spot, leftover

bindings/cpp/src/lib.rs
fresh-borzoni: Investigated this thoroughly. The current behavior matches Rust's InternalRow contract: callers must check is_null_at() before calling typed getters. The same here with isNull() check.

Added examples to match Rust tests.

fresh-borzoni:
@leekeiabstraction @zhaohaidao Ty for the review!

Addressed comments. 
PTAL 🙏

zhaohaidao (commented):
@fresh-borzoni Thank you. Only left one comment.

zhaohaidao:
> @zhaohaidao Thanks for the review. Answered comment. PTAL 🙏

Thanks for your explanation. LGTM

PR #353, opened 2026-02-18
chore: make RowView shared_ptr

fresh-borzoni:
## Summary

ScanRecord in C++ held a raw pointer to the underlying scan data, making it a borrowed view that dangled when ScanRecords was destroyed. Users couldn't store or accumulate records across Poll() calls, unlike Rust (Arc<RecordBatch>)  where records are self-contained.                    
                                                                                                                                                                                                                                                     
 Fixed by wrapping the opaque ScanResultInner in shared_ptr with a custom deleter that calls rust::Box::from_raw(). ScanRecord is now a value type: copyable, storable, safe to accumulate. Zero-copy preserved (same FFI pointer path), no Rust-side changes needed.

Updated docs to replace borrowing warnings with value-type semantics and added an accumulation example to log-tables guide.

### DIFF
--- a/bindings/cpp/include/fluss.hpp
+++ b/bindings/cpp/include/fluss.hpp
@@ -617,15 +617,15 @@ class GenericRow {
 
 /// Read-only row view for scan results. Zero-copy access to string and bytes data.
 ///
-/// WARNING: RowView borrows from ScanRecords. It must not outlive the ScanRecords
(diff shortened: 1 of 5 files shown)


fresh-borzoni:
@leekeiabstraction @zhaohaidao PTAL 🙏

PR #381, opened 2026-02-27
perf:support streaming download

zhaohaidao:
## Summary
- Add configurable streaming remote-log download path.
- Keep non-streaming range-read path as fallback.

## Why
Range-loop reads on remote segments increase per-chunk request overhead and can reduce effective object-store throughput. Streaming download improves transfer shape and reduces remote-read overhead for this workload profile.

## What Changed
- `crates/fluss/src/config.rs`:
  - `scanner_remote_log_streaming_read` (default `true`)
  - `scanner_remote_log_streaming_read_concurrency` (default `4`)
- `crates/fluss/src/client/table/scanner.rs`: pass streaming knobs into `RemoteLogDownloader`.
- `crates/fluss/src/client/table/remote_log.rs`:
  - add streaming path (`reader_with().chunk().concurrent()`)
  - keep range-read path for A/B and rollback
- `bindings/cpp/src/lib.rs`: switch to `Config::default() + field overrides` to avoid missing new config fields.

## Benchmark
- Fixed workload:
  - `cargo run --release --example custom-consume-table -- --task LdsHomeFeedQH --hours-ago 10 --bucket-mode range --max-buckets 4`
  - `timeout 180s` per run
| Factor | (r/s) |
|---|---:|
| `streaming_read=true` (B0_v3) 39750.45 |
| `streaming_read=false` (F6) 17950.00 |

- Turning streaming off: **-54.84%** throughput

### DIFF
--- a/crates/fluss/src/config.rs
+++ b/crates/fluss/src/config.rs
@@ -31,10 +31,6 @@ const DEFAULT_WRITER_BATCH_TIMEOUT_MS: i64 = 100;
 
 const DEFAULT_ACKS: &str = "all";
 
-fn default_scanner_remote_log_read_concurrency() -> usize {
-    DEFAULT_SCANNER_REMOTE_LOG_READ_CONCURRENCY
-}
-
 /// Bucket assigner strategy for tables without bucket keys.
 /// Matches Java `client.writer.bucket.no-key-assigner`.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, ValueEnum, Deserialize, Serialize)]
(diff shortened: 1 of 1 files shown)


luoyuxia (commented):
@zhaohaidao Thanks. LGTM overall. 
- also update bindings/python to respect the config
- update the docs in website directory

bindings/cpp/include/fluss.hpp
```diff
@@ -985,6 +985,10 @@ struct Configuration {
     size_t scanner_remote_log_prefetch_num{4};
     // Number of threads for downloading remote log data
     size_t remote_file_download_thread_num{3};
+    // Whether to use streaming reads for remote log download
+    bool scanner_remote_log_streaming_read{true};
+    // Streaming read concurrency within one remote log file
+    size_t scanner_remote_log_streaming_read_concurrency{4};
```
luoyuxia:
```suggestion
    size_t scanner_remote_log_read_concurrency{4};
```

luoyuxia (approved):
Thanks. LGTM! Will wait to see if there's more comments before merge

PR #392, opened 2026-02-28
chore:fix test_get_server_nodes to use shared cluster

fresh-borzoni:
as per title


fresh-borzoni:
@luoyuxia this should fix ci/cd

PR #396, opened 2026-02-28
Update API Ref following SASL / NoKeyAssigner PR / streaming file download

leekeiabstraction:
### Purpose

Update API Ref following SASL / NoKeyAssigner PR / streaming file download

### DIFF
--- a/website/docs/user-guide/cpp/api-reference.md
+++ b/website/docs/user-guide/cpp/api-reference.md
@@ -15,16 +15,24 @@ Complete API reference for the Fluss C++ client.
 
 ## `Configuration`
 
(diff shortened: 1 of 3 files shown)


leekeiabstraction:
@fresh-borzoni @zhaohaidao @charlesdong1991 PTAL 🙏

charlesdong1991 (approved):
Nice! Thanks for keeping docs up-to-date! 👍

charlesdong1991:
ohhh, i think since you are adding API reference, i think it's also nice to update python type stub in `python/fluss/__init__.pyi`, as that file can be useful for static type checking or autocomplete

Otherwise, i can throw a small PR to add to have this PR to focus on doc update! ^^

PR #404, opened 2026-03-01
[TASK-389] Idempotent writes

### LINKED ISSUE #389: Idempotent writes with writer ID allocation and batch sequencing
### Search before asking

### Description

The client retries failed batches on network errors up to i32::MAX times by default. Without idempotence, the server has no way to detect retried batches, when writer_id is -1 (current default), the server skips all dedup checks and appends unconditionally. A timeout where the server received the batch but the ack was lost results in silent duplicate records.

### Willingness to contribute

fresh-borzoni:
## Summary
closes #389

Implements idempotent writes for the Rust SDK. When enabled (default), the server allocates a writer ID and tracks per-bucket sequence numbers so it can detect and deduplicate retried batches. This makes retries safe - a transient network failure or leader failover no longer risks duplicate records.

## How it works

Before the first batch is sent, the sender requests a writer ID from the server via a new `InitWriter` RPC. Each batch is then stamped with `(writer_id, batch_sequence)` before being sent. The server uses this pair to reject duplicates: if it sees the same sequence again, it returns the previous result instead of writing twice.

The `IdempotenceManager` owns per-bucket state: the next sequence to assign, the in-flight batch queue, and error recovery logic. When a batch completes successfully, it's removed from the in-flight queue and the next sequence advances. When a batch fails with a retriable error, it gets re-enqueued with its original sequence number -  the server expects that exact sequence on retry.

Two server errors trigger a full reset of writer state: `UnknownWriterId` (server lost our writer session) and `OutOfOrderSequence` that can't be resolved by reordering. On reset, the writer ID is cleared, all pending sequences are invalidated, and a new writer ID is allocated on the next drain cycle. This matches Java's recovery semantics.

## Retry scenarios and sequence reordering

With up to 5 batches in-flight per bucket, responses can arrive out of order. The idempotence manager handles this by tracking per-bucket state: `last_acked_sequence`, an ordered in-flight queue, and `next_sequence`.

**Happy path - out-of-order completion.** Batches with sequences 0, 1, 2 are in-flight. Batch 1 completes before batch 0. `last_acked_sequence` advances to 1 (it always tracks the highest completed sequence, not strictly sequential). When batch 0 later completes, `last_acked_sequence` stays at 1 since 0 < 1. The server accepts both because they each carry the correct sequence. No reordering needed on the client.

**Retriable failure with sequence adjustment.** Batches 0, 1, 2 are in-flight. Batch 1 fails with a retriable error (e.g., `NetworkException`) and will be retried. But the server has never seen sequence 1, so when batch 1 is re-enqueued, there's a gap: the server expects 0, 1, 2 in order but would see 0, 2 (success), then 1 (retry). To fix this, `handle_failed_batch` with `adjust_sequences=true` removes batch 1 from the in-flight queue, decrements batch 2's sequence from 2 -> 1, and rolls back `next_sequence` from 3 -> 2. Now batch 2 arrives with sequence 1 (filling the gap), and when batch 1 is retried after re-drain, it gets a fresh sequence. The adjusted batches are marked with a `reset_batch_id` flag so `can_retry_for_error` knows they were rewritten and treats `OutOfOrderSequence` as retriable for them.

**Non-retriable failure - no adjustment.** If batch 1 exhausts its retries, `adjust_sequences=false`. Sequences are left as-is. The server will see a gap (sequence 1 missing), which triggers `OutOfOrderSequence` for batch 2. Since batch 2 is not the "next expected" sequence (`last_acked + 1`), `can_retry_for_error` returns true, it's a reordering issue, not a genuine violation. Batch 2 gets retried. If it *is* the next expected sequence and still gets `OutOfOrderSequence`, that's a genuine server-side inconsistency, the writer ID is reset and all state is cleared.

**`UnknownWriterId` / `OutOfOrderSequence` - full reset.** These indicate the server lost our session or detected an unrecoverable gap. `handle_failed_batch` clears the writer ID and wipes all bucket entries. On the next drain cycle, the sender allocates a fresh writer ID, and `maybe_update_writer_id` resets each bucket's sequence to 0 (but only once its in-flight queue is empty, to avoid corrupting batches still awaiting responses from the old session).

## Memory backpressure

The accumulator now enforces a memory budget (default 64MB) across all buffered batches. When the buffer is full, `upsert()`/`append()` block until in-flight batches complete and release memory. This prevents unbounded memory growth when the application writes faster than the network can drain. The `MemoryLimiter` is semaphore-style with byte granularity and RAII permits, memory is released automatically when a batch is completed or dropped.

## Where we diverged from Java
**Sequence assignment timing.** Java assigns sequences at `append()` time, its sender thread blocks at startup until the writer ID is allocated, so the ID is always available. Our `append()` is sync on the caller's thread with no guarantee the async writer ID allocation has completed yet, so we assign at `drain()` time instead as the sender has already ensured the ID exists before draining. E.g. the sender calls `maybe_wait_for_writer_id()` before each drain cycle, which allocates the ID via `InitWriter` RPC if needed, guaranteeing it exists before any batch is stamped.

**Sender concurrency model.** Java's sender calls `sendWriteData()` in a fire-and-forget style, with Netty I/O threads handling responses asynchronously. We use a `FuturesUnordered` event loop in a single tokio task: drain batches, fire all per-leader RPCs concurrently, then interleave response handling with the next drain cycle. This gives the same parallelism without dedicated threads per connection. When one leader is slow, responses from faster leaders still flow in and trigger new drain cycles, e.g no head-of-line blocking.

**Shutdown.** Java uses a `volatile boolean forceClose` flag that the sender loop polls. Tokio tasks support external cancellation, so we use `tokio::select!` with a timeout: signal graceful drain, wait up to the timeout for the sender to finish, then `JoinHandle::abort()` if it hasn't. `BroadcastOnce::Drop` automatically notifies any callers awaiting write results when their in-flight futures are cancelled, so no writes silently disappear.

### DIFF
--- a/crates/fluss/src/client/write/accumulator.rs
+++ b/crates/fluss/src/client/write/accumulator.rs
@@ -679,7 +679,7 @@ impl RecordAccumulator {
         if self.idempotence_manager.is_enabled()
             && ready_write_batch.write_batch.has_batch_sequence()
         {
-            if let Some(adjusted_seq) = self.idempotence_manager.take_adjusted_sequence(
+            if let Some(adjusted_seq) = self.idempotence_manager.get_adjusted_sequence(
                 &ready_write_batch.table_bucket,
                 ready_write_batch.write_batch.batch_id(),
             ) {
--- a/bindings/cpp/include/fluss.hpp
+++ b/bindings/cpp/include/fluss.hpp
@@ -1028,7 +1028,7 @@ struct Configuration {
     // Total memory available for buffering write batches (default 64MB)
     size_t writer_buffer_memory_size{64 * 1024 * 1024};
     // Maximum time in milliseconds to block waiting for buffer memory
-    uint64_t writer_buffer_wait_timeout_ms{60000};
+    uint64_t writer_buffer_wait_timeout_ms{std::numeric_limits<uint64_t>::max()};
     // Connect timeout in milliseconds for TCP transport connect
     uint64_t connect_timeout_ms{120000};
     // Security protocol: "PLAINTEXT" (default, no auth) or "sasl" (SASL auth)
--- a/crates/fluss/src/client/write/idempotence.rs
+++ b/crates/fluss/src/client/write/idempotence.rs
@@ -285,15 +285,18 @@ impl IdempotenceManager {
             .is_some_and(|b| b.batch_id == batch_id)
     }
 
-    /// Returns the current (possibly adjusted) in-flight sequence for a batch
-    /// and clears its reset flag in a single lock scope. Used by `re_enqueue`
-    /// to sync the WriteBatch's sequence with the adjusted InFlightBatch
-    /// sequence. Clearing the reset flag matches Java where `close()` resets
-    /// `reopened = false` — the hint is consumed after one re-enqueue cycle.
-    pub fn take_adjusted_sequence(&self, bucket: &TableBucket, batch_id: i64) -> Option<i32> {
-        let mut entries = self.bucket_entries.lock();
-        let entry = entries.get_mut(bucket)?;
-        entry.reset_batch_ids.remove(&batch_id);
+    /// Returns the current (possibly adjusted) in-flight sequence for a batch.
+    /// Used by `re_enqueue` to sync the WriteBatch's sequence with the adjusted
+    /// InFlightBatch sequence.
+    ///
+    /// Does NOT clear `reset_batch_ids` — the reset marker must survive
+    /// re-enqueue so that `can_retry_for_error` can still see it on subsequent
+    /// retries. It is cleared only on terminal events: `handle_completed_batch`
+    /// or `handle_failed_batch`. This matches Java where `reopened` persists
+    /// across retries and is only cleared in `close()` (resource cleanup).
+    pub fn get_adjusted_sequence(&self, bucket: &TableBucket, batch_id: i64) -> Option<i32> {
+        let entries = self.bucket_entries.lock();
+        let entry = entries.get(bucket)?;
         entry
             .in_flight
             .iter()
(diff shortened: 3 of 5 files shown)


fresh-borzoni:

I did my best with PR description to explain the feature, scenarios and why rust differs in some areas.

I'll handle docs in separate PR, they usually cause annoying conflicts, so I'd better avoid it in the same PR

crates/fluss/src/client/write/accumulator.rs
```diff
@@ -15,21 +15,142 @@
...
+
+/// Byte-counting semaphore that blocks producers when total buffered memory
+/// exceeds the configured limit. Matches Java's `LazyMemorySegmentPool` behavior.
+///
+/// TODO: Replace `notify_all()` with per-waiter FIFO signaling (Java uses per-request
+/// Condition objects in a Deque) to avoid thundering herd under high contention.
+///
+/// TODO: Track actual batch memory usage instead of reserving a fixed `writer_batch_size`
+/// per batch. This over-counts when batches don't fill completely, reducing effective
+/// throughput. Requires tighter coupling with batch internals.
```
leekeiabstraction: I think there's an undercounting problem as well i.e. if rows are large in bytes size

crates/fluss/src/client/write/accumulator.rs
```diff
@@ -15,21 +15,142 @@
...
+            wait_timeout,
+            closed: AtomicBool::new(false),
+            waiting_count: AtomicUsize::new(0),
+        }
+    }
+
+    /// Try to acquire `size` bytes. Blocks until memory is available,
+    /// the timeout expires, or the limiter is closed.
+    /// Returns a `MemoryPermit` on success.
+    pub fn acquire(self: &Arc<Self>, size: usize) -> Result<MemoryPermit> {
```
leekeiabstraction: Given this backpressure mechanism, would a hot bucket cause backpressure such that writes to other free buckets are slowed down or blocked?

luoyuxia (commented):
@fresh-borzoni Thanks for the greate pr. Overall LGTM, Left minor comment.

fresh-borzoni:
@luoyuxia I checked the run, these were errors pulling images from docker hub, both zookeeper:3.9.2 and apache/fluss:0.9.0-incubating failed
So not related to the changes, but to network transient issue

PR #415, opened 2026-03-03
[doc] Update fluss-rust readme markdown

luoyuxia:
### Purpose

Linked issue: close #xxx

See https://staging.crates.io/crates/fluss-rs-readme-preview
to preview the page in crate

### DIFF
--- a/bindings/python/pyproject.toml
+++ b/bindings/python/pyproject.toml
@@ -42,7 +42,7 @@ dependencies = [
 ]
 
 [project.urls]
-Homepage = "https://clients.fluss.apache.org/"
+Homepage = "https://clients.fluss.apache.org/user-guide/rust/installation/"
 Repository = "https://github.com/apache/fluss-rust"
 
 [project.optional-dependencies]
--- a/crates/fluss/Cargo.toml
+++ b/crates/fluss/Cargo.toml
@@ -23,7 +23,7 @@ version = { workspace = true }
 name = "fluss-rs"
 authors = { workspace = true }
 description = "The official rust client of Apache Fluss (Incubating)"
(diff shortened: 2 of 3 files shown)


luoyuxia:
@fresh-borzoni @leekeiabstraction Could you please help review this?

crates/fluss/README.md
```diff
@@ -1,25 +1,105 @@
...
 
-## Requirements
+## Usage
 
-- Rust (see [rust-toolchain.toml](../../rust-toolchain.toml) at repo root)
-- protobuf (for build)
+The following example shows both **primary key (KV) tables** and **log tables** in one flow: connect, create a KV table (upsert + lookup), then create a log table (append + scan).
 
-## Build
+```rust
```
charlesdong1991: i wonder how easy it is to maintain this code snippet long-term in README.md, since it's not compiled, so API changes will silently make it stale. We already have examples in crates/examples/ that cover these flows. 

Would we consider to keep short quickstart snippet to just give users a feel for API, and link example folder for detailed workflows. This was inspired by what OpenDAL does: https://github.com/apache/opendal/blob/main/core/README.md

luoyuxia: Thanks for the thoughtful discussion! I think we can keep the current approach for now — 
the snippet is relatively small, and the API it demonstrates is stable enough that it's 
unlikely to go stale anytime soon. Let's leave it as-is and revisit if it becomes a 
real maintenance pain point down the road.

charlesdong1991: cool! Let's keep as is and revisit if becoming an issue long term! Thanks for the discussion!

PR #420, opened 2026-03-03
chore: fix mut schema builder public api

fresh-borzoni:
Fix mut requirement for schema builder


fresh-borzoni:
@luoyuxia I guess with this we don't need users use `mut`
I checked, we can leave callers with `mut` references for reassignment for internal code. but for public API we shouldn't require users to use mutable schema

fresh-borzoni:
fixed clippy warning about not used `mut`

PR #437, opened 2026-03-08
chore: Check tablet server availabitility in cpp tests

charlesdong1991:
i think tests in cpp fail now, and we should check tablet server availability as well?


charlesdong1991:
i think this should fix, but PTAL @fresh-borzoni @luoyuxia 🙏

fresh-borzoni (approved):
@charlesdong1991 Ty for the PR, LGTM

Q: have you experienced cpp integration tests failure?
In practice, all processes would try quick check and fail and go to WaitForCluster() which checks all ports. 
So in order for tests to fail, one process should start docker and have coordinator port ready, when other processes are still on a fast check. As all the processes start quickly and go for 1s fast check - close to zero chances practically.

But it's fine as correctness fix 👍

luoyuxia (approved):
@charlesdong1991 +1

PR #438, opened 2026-03-09
feat: add async 'for' loop support to LogScanner (#424)

### LINKED ISSUE #424: Python: add async 'for' loop support to LogScanner
### Search before asking

### Description

LogScanner should implement __aiter__/__anext__ so users can write smth like this:
```python
  async for record in scanner:
      process(record)
```
  Instead of the current manual poll loop:
```python
  while True:
      records = scanner.poll(1000)
      for record in records:
          process(record)
```

the same for record_batch_log_scanner and log_scanner, they differ in iterable types, but overall that's the idea

### Willingness to contribute

qzyu999:
### Purpose

Linked issue: close #424 

This pull request completes Issue #424 by enabling standard cross-boundary native Python `async for` language built-ins over the high-performance PyO3 wrapped `LogScanner` stream instance.

### Brief change log

Previously, PyFluss developers had to manually orchestrate `while True` polling loops over network boundaries using `scanner.poll(timeout)`. This PR refactors the Python `LogScanner` iterator logic by implementing the async traversal natively via Rust `__anext__` polling bindings and Python Generator `__aiter__` context adapters:

- **State Independence**: Refactored `ScannerKind` internals into a safely buffered `Arc<tokio::sync::Mutex<ScannerState>>`. This guarantees strict thread-safety and fulfills Rust's lifetime constraints enabling unboxed state transitions inside the `python_async_runtimes` `tokio` closure.
- **Asynchronous Execution**: Polling evaluates non-blocking loops. PyFluss automatically maps Arrow records onto the `.await` future yield sequence smoothly without blocking event cycles or hardware threads directly!
- **Iterable Compliance**: To correctly resolve runtime `inspect.isasyncgen()` compliance checks within strictly versioned Python 3.12+ engines (such as modern IPython Jupyter servers), `__aiter__` dynamically generates a properly wrapped coroutine generator dynamically inside the codebase via `py.run()`. This completely masks the Python ecosystem's iterator type limitations automatically out-of-the-box.

### Tests

- [NEW] `test_log_table.py::test_async_iterator`: Integrated a testcontainers ecosystem confirming zero-configuration iteration capabilities function natively evaluating `async for record in scanner` perfectly without pipeline interruptions while yielding thousands of appended instances sequentially backwards matching existing legacy data frameworks.

### API and Format

Yes, this expands the API natively extending capabilities allowing `async for` loops gracefully. Existing user logic leveraging explicit implementations of `.poll_arrow()` or legacy functions are untouched.

### Documentation

Yes, I updated integration tests acting as live documentation proof demonstrating the capability natively.


qzyu999:
Hi @fresh-borzoni, thanks for the recommendations. I've taken a look and came up with the following changes, PTAL when available:

1. Replace `Arc<tokio::sync::Mutex<ScannerState>>` with `Arc<ScannerKind>`
2. Remove `ScannerState` struct and `VecDeque` buffer
3. Remove all 6 `unsafe` pointer casts
4. Replace `__anext__` with `_async_poll(timeout_ms)` (single bounded poll)
5. Replace `__aiter__` with `PyOnceLock`-cached Python async generator
6. Change batch scanner error from `StopAsyncIteration` to `TypeError`
7. Update `with_scanner!` macro or inline to use `&self.kind` directly
8. Add break-safety and batch-scanner-error tests

bindings/python/src/table.rs
```diff
@@ -2199,6 +2200,171 @@ impl LogScanner {
         Ok(df)
     }
 
+    fn __aiter__<'py>(slf: PyRef<'py, Self>) -> PyResult<Bound<'py, PyAny>> {
```
fresh-borzoni: we need to add this method to .pyi stubs

fresh-borzoni (commented):
@qzyu999 TY, looked through, left comments, PTAL

bindings/python/src/table.rs
qzyu999: Hi @fresh-borzoni, just added `__aiter__` to `__init__.pyi` here 134e56b543395d7f110ffa7b36edfe6fdab4e729 along with with `_async_poll` and `_async_poll_batches`.

fresh-borzoni (commented):
@qzyu999 Ty, nice work, only nit comments remain, PTAL

fresh-borzoni (approved):
@qzyu999 Ty, LGTM

bindings/python/src/table.rs
```diff
@@ -2198,6 +2197,156 @@ impl LogScanner {
...
+    ///
+    /// Args:
+    ///     timeout_ms: Timeout in milliseconds for the network poll (default: 1000)
+    ///
+    /// Returns:
+    ///     Awaitable that resolves to a list of ScanRecord objects
+    fn _async_poll<'py>(
+        &self,
+        py: Python<'py>,
+        timeout_ms: Option<i64>,
```
leekeiabstraction: It seems like we hardcode this anyway in generated code, do we need to define an arg for this? 

Also nit: poll_interval_ms is more accurate

bindings/python/src/table.rs
```diff
@@ -2198,6 +2197,156 @@ impl LogScanner {
...
+    ) -> PyResult<Bound<'py, PyAny>> {
+        let timeout_ms = timeout_ms.unwrap_or(1000);
```
leekeiabstraction: nit: magic number

bindings/python/src/table.rs
```diff
@@ -2198,6 +2197,156 @@ impl LogScanner {
...
+    ///     timeout_ms: Timeout in milliseconds for the network poll (default: 1000)
+    ///
+    /// Returns:
+    ///     Awaitable that resolves to a list of RecordBatch objects
+    fn _async_poll_batches<'py>(
+        &self,
+        py: Python<'py>,
+        timeout_ms: Option<i64>,
+    ) -> PyResult<Bound<'py, PyAny>> {
+        let timeout_ms = timeout_ms.unwrap_or(1000);
```
leekeiabstraction: nit: magic number

leekeiabstraction (commented):
TY for the PR. Left a few minor comments, PTAL.

bindings/python/src/table.rs
qzyu999: Hi @leekeiabstraction, in d52113357d272192d8262f1de2b25006d50fb1ac I made this into a global variable, `DEFAULT_POLL_INTERVAL_MS` that is initialized (to 1000) at the top of the script. It's then later referenced within the `_async_poll` and `_async_poll_batches` functions.

bindings/python/src/table.rs
qzyu999: Hi @leekeiabstraction, this is mentioned above.

bindings/python/src/table.rs
qzyu999: Hi @leekeiabstraction, this is mentioned above.

qzyu999:
> TY for the PR. Left a few minor comments, PTAL.

Hi @leekeiabstraction, I updated the code to use a global `DEFAULT_POLL_INTERVAL_MS=1000` at the top, rather than have the `timeout_ms=1000` as a magic number within the function, where `DEFAULT_POLL_INTERVAL_MS` is then referenced within the function. I agree that the magic number issue isn't ideal and I believe this is a suitable fix.

In an unrelated change, I noticed that the code was no longer compiling when running `uv run maturin develop`, seems like the following code:
```python
#[pyclass]
pub struct LogScanner {
  ...
      admin: fcore::client::FlussAdmin,
```
requires being updated to
```python
#[pyclass]
pub struct LogScanner {
  ...
      admin: Arc<fcore::client::FlussAdmin>,
```
in order to compile properly. I believe there were some changes on the `main` branch that required this to happen. Looking at the git history, it seems that in 3bd6b1b392b767c2760e95db671ef6aa90e3bafa (I saw some later changes to this also, but this is the breaking change), the update was made to change (in `connection.rs`):
```rust
pub async fn get_admin(&self) -> Result<FlussAdmin> {
```
to
```rust
pub async fn get_admin(&self) -> Result<Arc<FlussAdmin>> {
```

PTAL and let me know what you think about these solutions, thanks!

bindings/python/test/test_log_table.py
```diff
@@ -729,6 +729,382 @@ async def test_scan_records_indexing_and_slicing(connection, admin):
...
+        f"Unexpected IDs: {all_ids - set(range(1, 11))}"
+    )
+
+    await admin.drop_table(table_path, ignore_if_not_exists=False)
+
+
+async def test_async_iterator_multiple_batches(connection, admin):
+    """Verify async iteration works across multiple network poll cycles.
+
+    _async_poll does a single bounded poll per call.  Writing 20 records
```
leekeiabstraction, line 859: Q: how does 20 trigger multiple batches?

leekeiabstraction (approved):
LGTM. TY for your contribution!

bindings/python/test/test_log_table.py
qzyu999: Hi @leekeiabstraction, good question, perhaps the docstring wasn't clear enough. The idea is that since we initialize the table with 3 buckets, by writing 20 records we ensure that the data is statistically distributed across multiple buckets. So when we run  _async_poll, we're guaranteed to loop through several calls for each bucket to collect the batches of data.

So the test here verifies that the async iterator can correctly handle data arriving in multiple chunks over time without losing records or leaking resources.

Please let me know if any further clarification is needed for the intent of this test. Some updates can be made if necessary.

PR #442, opened 2026-03-15
Feat : Row data type support in Rust

### LINKED ISSUE #388: Row data type support in Rust
### Search before asking

### Description

Add end-to-end support for ROW (nested struct) columns. 
A nested Row is just a recursively-embedded CompactedRow, the same format as a top-level row, length-prefixed. This means the existing CompactedRowWriter/CompactedRowReader logic is reused recursively, making this simpler than it sounds.

The Datum enum needs a Row variant, InternalRow needs get_row(). Look at Java's CompactedRowWriter.java (lines 339-346 atm) and CompactedRowReader.java (lines 370-378 atm) for how nested rows are written and read - it's serialize-to-bytes then write_bytes, same as strings. The value writer, field getter, and key encoder need the same additions as the other two types.

Tests should cover simple nested rows, deeply nested rows (Row inside Row), rows with nullable fields, and once all three compex data types land, combination tests like a Row containing Array and Map fields. May also be done as the separate issue.

Java reference: CompactedRowWriter.java , CompactedRowReader.java

### Willingness to contribute

hemanthsavasere:
### Purpose

Linked issue: close #388

DataType Row was defined in the schema layer but the serialization stack was entirely missing. CompactedRowWriter and Reader panicked, FieldGetter and ValueWriter hit unimplemented macros, and Datum had no Row variant. This PR wires up the full stack.

### Brief change log

- Add Datum Row variant and as_row accessor.
- Add get_row to InternalRow trait with a default error implementation.
- Implement get_row on GenericRow and CompactedRow.
- Implement ColumnarRow get_row via Arrow StructArray extraction with OnceLock caching per column, invalidated on set_row_id.
- Add InnerValueWriter Row which serializes into a temporary CompactedRowWriter then calls write_bytes, matching the Java wire format.
- Add DataType Row arm in CompactedRowDeserializer using read_bytes and recursive deserialize.
- Add InnerFieldGetter Row which automatically enables ROW in CompactedKeyEncoder.
- Handle Datum Row in C++ resolve_row_types.

### Tests

- test_row_simple_nesting: round-trip ROW of INT and STRING.
- test_row_deep_nesting: round-trip ROW of ROW of INT.
- test_row_with_nullable_fields: null field inside nested row and null outer ROW column.
- test_row_as_primary_key: ROW through CompactedKeyEncoder; asserts non-empty, deterministic, and distinguishable output.
- columnar_row_reads_nested_row, columnar_row_reads_deeply_nested_row, and columnar_row_get_row_cache_invalidated_on_set_row_id for Arrow StructArray extraction.

### API and Format

API: Datum Row is a new variant. Exhaustive match on Datum will require a new arm. InternalRow get_row is additive with a default implementation.

Wire format: Unchanged. ROW uses varint-length plus CompactedRow blob, identical to String or Bytes. This matches the Java reference byte-for-byte.

### Documentation

No new documentation needed.


crates/fluss/src/row/column.rs
```diff
@@ -533,4 +715,112 @@ mod tests {
             .unwrap()
         );
     }
+
+    fn make_struct_batch(
```
leekeiabstraction: Is this for test only? If so, move under '#[test]'/mod?

leekeiabstraction (commented):
Thank you for your contribution! Left some comments/questions, PTAL.

crates/fluss/src/row/binary/binary_writer.rs
```diff
@@ -237,6 +238,26 @@ impl InnerValueWriter {
             (InnerValueWriter::TimestampLtz(p), Datum::TimestampLtz(ts)) => {
                 writer.write_timestamp_ltz(ts, *p);
             }
+            (InnerValueWriter::Row(row_type), Datum::Row(inner_row)) => {
+                use crate::row::compacted::CompactedRowWriter;
+                let field_count = row_type.fields().len();
+                let mut nested = CompactedRowWriter::new(field_count);
+                for (i, field) in row_type.fields().iter().enumerate() {
+                    let datum = &inner_row.values[i];
```
charlesdong1991: potential panic on OOB?

crates/fluss/src/row/column.rs
```diff
@@ -407,16 +577,28 @@ impl InternalRow for ColumnarRow {
...
             .value(self.row_id))
     }
+
+    fn get_row(&self, pos: usize) -> Result<&GenericRow<'_>> {
+        let lock = self.nested_rows.get(pos).ok_or_else(|| IllegalArgument {
+            message: format!("column index {pos} out of bounds for get_row"),
+        })?;
+        let batch = Arc::clone(&self.record_batch);
+        let row_id = self.row_id;
+        Ok(lock.get_or_init(|| {
```
charlesdong1991: maybe better to use `get_or_try_init` here?

charlesdong1991 (commented):
thanks for the PR! left couple comments

leekeiabstraction:
@hemanthsavasere Bump, are you able to address the comments? Thank you

fresh-borzoni:
Reviving this as #544 - rebased on current main, addressed review feedback, fixes and improvements.
 @hemanthsavasere your commit is preserved with original authorship, thanks for the original implementation 🙇

PR #451, opened 2026-03-24
[BUG] Stale cluster state in Lookuper lookup

fresh-borzoni:
`Lookuper::lookup()` captures a cluster state before calling `leader_for()`, then uses that state to resolve the leader's server. If `leader_for()` refreshed metadata internally (cache miss -> coordinator RPC), the returned leader could reference a server absent from the stale state, causing a spurious "server not found" error during leader transitions.

Proposed fix: connect to the ServerNode returned by `leader_for()` directly as it already contains host and port. 
Removes the redundant cluster state lookup and the race entirely.


leekeiabstraction (approved):
TY for the PR!

PR #474, opened 2026-04-01
Add array data type support for Python

### LINKED ISSUE #469: [python] Array data type support
### Search before asking

### Description

Add array data type support for python. Rust SDK core and client wiring is here: https://github.com/apache/fluss-rust/pull/433

### Willingness to contribute

qzyu999:
### Purpose

Linked issue: close #469 

The purpose of this change is to complete the Python implementation for Array types by adding support for deterministic-length arrays (FixedSizeList). This ensures the Python client can interface with all standard Arrow list layouts while providing an idiomatic, programmatic way to construct nested schemas.

### Brief change log
* **Core Engine Refinement (`crates/fluss/src/row/column.rs`):** Added explicit downcasting for `FixedSizeListArray`. This preserves performance by calculating element positions via direct multiplication, avoiding the memory fetch overhead of an offsets buffer.
* **Programmatic Schema API (`bindings/python/src/metadata.rs`):** Introduced a `DataTypes` factory class for Python. This enables the construction of nested types (e.g., `DataTypes.array(DataTypes.int())`) which was previously impossible for FFI-backed types.
* **Idiomatic FFI Bindings:** Replaced inherent `.to_string()` methods with the standard Rust `fmt::Display` trait for `DataType`. This standardizes string representation across Rust and Python (`__str__` and `__repr__`). This should also be reusable for when adding Map<Key, Value> or Struct/Row.
* **Arrow Translator Update (`crates/fluss/src/record/arrow.rs`):** Updated the type translator to unify `List`, `LargeList`, and `FixedSizeList` into a single logical Fluss `Array` type.
* **Linter & Precision Pass:** Cleaned up `clippy::clone_on_copy` warnings in integration tests and replaced hardcoded float approximations with `std::f32::consts::PI` and `std::f64::consts::E` to prevent precision drift in streaming aggregations.

### Tests
* **Rust Unit Tests:** Added `test_from_arrow_type_fixed_size_list` to verify the Arrow-to-Fluss type translation.
* **Python Metadata Tests:** Added tests in `test_schema.py` to verify the `DataTypes` factory and string representation logic.
* **Integration Tests:** * `test_append_and_scan_with_array`: Verifies round-trip for variable-length arrays.
    * `test_append_and_scan_with_fixed_size_array`: Verified client-side but currently **SKIPPED** in CI. This test requires a Fluss server version >= 0.9.1 to handle the new storage layout.

### API and Format
* **API:** This change adds a new `DataTypes` factory to the Python API. It also improves the `__repr__` output for schemas, making nested types human-readable.
* **Format:** This PR introduces support for the `FixedSizeList` Arrow storage format within the Fluss engine, which is more space-efficient and performant for fixed-length vector data (like coordinates or embeddings).

### Documentation
This change introduces a new feature: Support for the `Array` data type in the Python client, including support for `FixedSizeList`. Users can now define, write, and read array-based columns using the Python SDK.


crates/fluss/src/row/binary_array.rs
```diff
@@ -621,6 +621,79 @@ impl FlussArrayWriter {
...
+    fn get_bytes(&self, pos: usize) -> Result<&[u8]> {
+        self.get_binary(pos)
+    }
+
+    fn get_array(&self, pos: usize) -> Result<FlussArray> {
+        self.get_array(pos)
+    }
+
+    fn as_encoded_bytes(&self, _write_format: crate::client::WriteFormat) -> Option<&[u8]> {
+        Some(self.as_bytes())
```
fresh-borzoni: I don't think it's encoded

bindings/python/src/table.rs
```diff
@@ -1370,6 +1370,20 @@ pub fn datum_to_python_value(
                 .map_err(|e| FlussError::from_core_error(&e))?;
             rust_timestamp_ltz_to_python(py, ts)
         }
+        DataType::Array(array_type) => {
```
fresh-borzoni: what about python_value_to_datum?

bindings/python/src/metadata.rs
```diff
@@ -765,3 +766,116 @@ impl DatabaseInfo {
         Self { __info: info }
     }
 }
+
+/// Represents a Fluss data type
+#[pyclass]
+#[derive(Clone)]
+pub struct DataType {
```
fresh-borzoni: Is it even used aside from tests?

fresh-borzoni (commented):
@qzyu999 Ty for the PR, left comments, PTAL

crates/fluss/src/row/binary_array.rs
qzyu999: Hi @fresh-borzoni, thanks for pointing this out. I just removed it in d812d393d4debb0bbf7b41bc1eb5b317819974c1 so that it reverts to the default behavior of `InternalRow` by returning `None` if this function is called.

bindings/python/test/test_log_table.py
```diff
@@ -755,3 +756,98 @@ def _poll_arrow_ids(scanner, expected_count, timeout_s=10):
...
+    # Verify via to_arrow (batch-based)
+    scanner2 = await table.new_scan().create_record_batch_log_scanner()
+    scanner2.subscribe_buckets({0: fluss.EARLIEST_OFFSET})
+    result_table = scanner2.to_arrow()
+
+    assert result_table.num_rows == 2
+    assert result_table.column("tags").to_pylist() == [["a", "b"], ["c"]]
+    assert result_table.column("scores").to_pylist() == [[10, 20], [30]]
+
+@pytest.mark.skip(reason="FixedSizeList support requires server-side updates (≥0.9.1). "
```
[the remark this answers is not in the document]
qzyu999: Hi @fresh-borzoni, thanks for asking, I removed the `pytest.mark.skip()` in 9071cf82df02ad647a6075ede40ace2c22c74ca4. The reason it's there is because I was testing it locally and was having some issues and wasn't sure of what to do. My understanding was that the test would fail in GitHub's CI due to something about the server-side not being able to test `FixedSizeList` as you mentioned. I thought I would just include the test since I had confirmed locally that it worked.

Edit: I thought about it some more and just updated the test to have a proper message in the `reason` message in 87ab426a42a1daa46bc50e1e5e1fe72295e7b8e5. When testing locally I had to keep adding it back so the tests would pass without the error message. I think we should keep the test itself as my understanding is that `FixedSizeList` is important to adding support for arrays. I think we can leave it here and remove it once the main repo has support for `FixedSizeList`. Otherwise, I can remove this `test_append_and_scan_with_fixed_size_array` test entirely. Please let me know what you think makes the most sense.

bindings/python/src/table.rs
qzyu999: Hi @fresh-borzoni, thanks for catching this, added functionality for `python_value_to_datum` in 7dec9da4a9b8989766b08766b9521cbb01462509 along with a set of tests.

bindings/python/test/test_log_table.py
```diff
@@ -755,3 +756,98 @@ def _poll_arrow_ids(scanner, expected_count, timeout_s=10):
...
+    )
+    schema = fluss.Schema(pa_schema)
+    table_descriptor = fluss.TableDescriptor(schema)
+    await admin.create_table(table_path, table_descriptor, ignore_if_exists=False)
+
+    table = await connection.get_table(table_path)
+    append_writer = table.new_append().create_writer()
+
+    # Batch 1: Testing both standard and large lists
+    batch1 = pa.RecordBatch.from_arrays(
```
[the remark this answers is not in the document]
qzyu999: Hi @fresh-borzoni, added null tests here b16a9b0951e870e5acaffeab9347137c4e91e6ed.

bindings/python/src/lib.rs
```diff
@@ -130,6 +130,8 @@ fn _fluss(m: &Bound<'_, PyModule>) -> PyResult<()> {
     m.add_class::<WriteResultHandle>()?;
     m.add_class::<DatabaseDescriptor>()?;
     m.add_class::<DatabaseInfo>()?;
+    m.add_class::<DataType>()?;
```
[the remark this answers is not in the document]
qzyu999: Hi @fresh-borzoni, added the .pyi stubs in 6a0fdf20db532a0ed1fca4fd1a4bfae0c8be059b.

Edit: Removed as of 5315521b89b4ffe1de53066d8fba8de88cb6b2ff (reason mentioned below).

bindings/python/src/metadata.rs
qzyu999: Hi @fresh-borzoni, this is actually a valid point, none of that `DataType` usage is useful anywhere outside of testing. I was looking to build a way to validate the nested data types within arrays, but it's never made available to the actual API itself since Fluss relies on PyArrow. I've removed all such references as of 5315521b89b4ffe1de53066d8fba8de88cb6b2ff.

crates/fluss/src/row/column.rs
```diff
@@ -607,29 +615,50 @@ fn write_arrow_values_to_fluss_array(
             )?;
         }
         DataType::Array(_) => {
-            let list_arr =
```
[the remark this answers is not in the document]
qzyu999: Hi @fresh-borzoni, thanks for catching this, I've added changes for this in c81de6355b5252303ee9bfccf0c3551ac3a1c59c. I've refactored the `DataType::Array(_)` arm to hoist both the `downcast_ref` and the `from_arrow_type` resolution entirely out of the element loop. To avoid duplicating the new hoisted loop for all three Arrow list types, I introduced a `write_list_elements!` macro. The type checking now correctly happens just once per array.

qzyu999:
Hi @fresh-borzoni, just completed going over your latest feedback, PTAL. Thanks!

crates/fluss/src/row/column_writer.rs
```diff
@@ -321,6 +327,26 @@ impl ColumnWriter {
                     }
                 }
             }
+            DataType::Array(array_type) => {
+                let element_type = array_type.get_element_type();
+                let arrow_element_type = match arrow_type {
+                    ArrowDataType::List(field) | ArrowDataType::LargeList(field) => {
```
leekeiabstraction: Curious, why only match against List / LargeList but not FixedSizeList?

If I understand correctly, there's no code path that would reach LargeList / FixedSizeList here so matching against LargeList is defensive but inconsistent in that we do not match against FixedSize

crates/fluss/src/row/column_writer.rs
```diff
@@ -550,15 +576,507 @@ impl ColumnWriter {
...
+                TimestampNanosecondBuilder::with_capacity(capacity),
+            )),
+            _ => Err(Error::IllegalArgument {
+                message: format!("Unsupported Arrow type for TimestampLTz: {arrow_type:?}"),
+            }),
+        },
+        DataType::Array(array_type) => {
+            let element_type = array_type.get_element_type();
+            let arrow_element_type = match arrow_type {
+                ArrowDataType::List(field) | ArrowDataType::LargeList(field) => field.data_type(),
```
leekeiabstraction: Curious, why only match against List / LargeList but not FixedSizeList?

If I understand correctly, there's no code path that would reach LargeList / FixedSizeList here so matching against LargeList is defensive but inconsistent in that we do not match against FixedSize

bindings/python/test/test_log_table.py
```diff
@@ -755,3 +754,244 @@ def _poll_arrow_ids(scanner, expected_count, timeout_s=10):
...
+        [10, 20],
+        [30],
+        [40, 50],
+        [60],
+        None,
+        [],
+    ]
+
+
+@pytest.mark.skip(reason="Server currently only accepts ListVector. FixedSizeList causes IPC mismatch until server supports it.")
```
leekeiabstraction: Do we need to have the same for LargeList? Or test (skipped) both LargeList and FixedSizeList in here

crates/fluss/tests/integration/kv_table.rs
```diff
@@ -632,8 +632,8 @@ mod kv_table_test {
         let col_smallint = 32767i16;
         let col_int = 2147483647i32;
         let col_bigint = 9223372036854775807i64;
-        let col_float = 3.14f32;
-        let col_double = 2.718281828459045f64;
+        let col_float = std::f32::consts::PI;
```
leekeiabstraction, line 635: These seems unrelated to Array data type support change?

crates/fluss/tests/integration/log_table.rs
```diff
@@ -738,18 +738,18 @@ mod table_test {
         row.set_field(14, col_time_ns);
         row.set_field(15, col_timestamp_s);
         row.set_field(16, col_timestamp_ms);
-        row.set_field(17, col_timestamp_us.clone());
-        row.set_field(18, col_timestamp_ns.clone());
+        row.set_field(17, col_timestamp_us);
```
leekeiabstraction, line 741: These seems unrelated to Array data type support change?

bindings/python/test/test_schema.py
```diff
@@ -35,3 +35,38 @@ def test_get_primary_keys():
     assert schema_without_pk.get_primary_keys() == []
 
 
+def test_schema_with_array():
```
leekeiabstraction, line 38: Do we need to update documentation as well on Array data type support?

leekeiabstraction (commented):
TY for the PR. Left some comments. PTAL

crates/fluss/src/row/column_writer.rs
qzyu999: Hi @leekeiabstraction, good catch, I agree I missed this here. Added in 6fcc0025f5ff4e688c1af3ba40154c889ec5b98d.

crates/fluss/src/row/column_writer.rs
qzyu999: Hi @leekeiabstraction, responded to this in the above reply.

crates/fluss/tests/integration/kv_table.rs
qzyu999: Hi @leekeiabstraction, I just thought these might be some minor fixes to do along the way, can remove if it doesn't look right.

crates/fluss/tests/integration/log_table.rs
qzyu999: Hi @leekeiabstraction, I just thought these might be some minor fixes to do along the way, can remove if it doesn't look right.

bindings/python/test/test_schema.py
qzyu999: Hi @leekeiabstraction, documentation updated in 5210fe57c448368b9aa2be223474c696ef8fb0d7 for PyArrow list, large list, and fixed size list data types.

bindings/python/test/test_log_table.py
qzyu999: Hi @leekeiabstraction, thanks for catching this, added a test for FixedSizeList in a2fdfb5a6113a7db25e0c289e25544647d71951f, but again it needs to be skipped via `pytest` also.

qzyu999:
> TY for the PR. Left some comments. PTAL

Hi @leekeiabstraction, I've addressed all the review comments, PTAL. For `kv_table.rs` and `log_table.rs`, I can simply revert those changes if that's simpler. Thanks!

leekeiabstraction:
@fresh-borzoni Any further comments on this PR?

fresh-borzoni:
@leekeiabstraction @qzyu999 Ty, I'll review this evening.

just quick glimpse: The Java Arrow library won't deserialize a FixedSizeList into a ListVector, that's why we needed to skip test, I don't think this is the right approach.

fresh-borzoni (commented):
@qzyu999

Thank you for the changes, I reviewed holistically, and I think we shall go this way:
* Drop FixedSizeList/LargeList. Server only accepts ListVector, so to support this we need server-side changes first, which is out of scope.
* Use nested ColumnWriter instead of ListBuilder<Box<dyn ArrayBuilder>>. Type erasure forces duplicating all builder creation in column_writer.rs.  Nested ColumnWriter, on the contrary, can reuse existing typed dispatch. e.g. store a Box<ColumnWriter> for elements inside the TypedWriter::List variant
smth like this:

``` rust
/// element_writer  - Box<ColumnWriter>
  TypedWriter::List { element_writer, offsets, non_null, .. } => {
      let array = row.get_array(pos)?;
      for i in 0..array.size() {
          if array.is_null_at(i) {
              element_writer.append_null();
          } else {
              element_writer.write_non_null_at(&array, i)?;
          }
      }
      offsets.push(i32::try_from(last + array.size())?);
      non_null.push(true);
      Ok(())
  }
```

cc @leekeiabstraction @luoyuxia

qzyu999:
> @qzyu999
> 
> Thank you for the changes, I reviewed holistically, and I think we shall go this way:
> 
> * Drop FixedSizeList/LargeList. Server only accepts ListVector, so to support this we need server-side changes first, which is out of scope.
> * Use nested ColumnWriter instead of ListBuilder<Box>. Type erasure forces duplicating all builder creation in column_writer.rs.  Nested ColumnWriter, on the contrary, can reuse existing typed dispatch. e.g. store a Box for elements inside the TypedWriter::List variant
>   smth like this:
> 
> ```rust
> /// element_writer  - Box<ColumnWriter>
>   TypedWriter::List { element_writer, offsets, non_null, .. } => {
>       let array = row.get_array(pos)?;
>       for i in 0..array.size() {
>           if array.is_null_at(i) {
>               element_writer.append_null();
>           } else {
>               element_writer.write_non_null_at(&array, i)?;
>           }
>       }
>       offsets.push(i32::try_from(last + array.size())?);
>       non_null.push(true);
>       Ok(())
>   }
> ```
> 
> cc @leekeiabstraction @luoyuxia

Hi @fresh-borzoni, thanks for the suggestions, I've made the changes in the latest set of commits: 068f210cf6ec881cb337731019beecf5089db41b, 7c6c3ea2aab49f7730fe411b27c4efa4526988d9, e0d961b99550c9012927017a677d0751e8c78d2b. 068f210cf6ec881cb337731019beecf5089db41b contains the main changes, others are removing of references to FixedSizeList and LargeList. Please let me know if any further changes are necessary.

Edit: Updated the commit hash after doing a rebase due to merge conflicts after updates to the main branch.

fresh-borzoni (approved):
@qzyu999 Ty, LGTM 👍

crates/fluss/src/row/column_writer.rs
```diff
@@ -339,41 +365,65 @@ impl ColumnWriter {
...
     pub fn finish(&mut self) -> ArrayRef {
-        self.as_builder_mut().finish()
+        match &mut self.inner {
+            TypedWriter::List {
+                element_writer,
+                offsets,
+                validity,
+            } => {
+                let values = element_writer.finish();
+                finish_list_array(values, offsets, validity)
```
leekeiabstraction: I think we need to reset offsets and validity vectors. If we optimise in the future to re-use column writer, subsequent finish() will operate on dirty offsets and validity. Something like the following?

```rust
let values = element_writer.finish();
let taken_offsets = std::mem::replace(offsets, vec![0]);                                                                                                                                                                                                                                               
let taken_validity = std::mem::take(validity);          
finish_list_array(values, &taken_offsets, &taken_validity)         
```

crates/fluss/src/row/column_writer.rs
```diff
@@ -550,15 +600,53 @@ impl ColumnWriter {
...
 
+fn finish_list_array(values: ArrayRef, offsets: &[i32], validity: &[bool]) -> ArrayRef {
+    use arrow::array::ListArray;
+    use arrow::buffer::{NullBuffer, OffsetBuffer, ScalarBuffer};
+    use arrow::datatypes::{Field, FieldRef};
+    use std::sync::Arc;
+
+    let offsets_buffer = OffsetBuffer::new(ScalarBuffer::from(offsets.to_vec()));
+    let null_buffer = NullBuffer::from(validity.to_vec());
+    let field = Arc::new(Field::new("item", values.data_type().clone(), true));
```
leekeiabstraction: It seems like we're hardcoding to be nullable. Will this cause an issue when user defines type to be non nullable and use batch builder?

If this is an actual issue, we should add test coverage as well.

leekeiabstraction (commented):
TY for the updates and apologies for the response time, I left a couple more comments.

There's a failed CI documentation check run as well. I had a quick look but do not immediately know how the doc changes could have caused that.

crates/fluss/src/row/column_writer.rs
qzyu999: Hi @leekeiabstraction, thanks for the clarification, the above changes were added in 8a6742c986e412f4ad8c4cd70164410bd230d8d6.

crates/fluss/src/row/column_writer.rs
qzyu999: Hi @leekeiabstraction, the hardcoding has been fixed in 8a6742c986e412f4ad8c4cd70164410bd230d8d6, and an upgraded test was added in e5267c3b00b0bf1fe988386265816e8384086b74. The `datatype.rs` was also updated in 60a14d1f33a78adf75a763d39348482e1b7493d5 to integrate `as_non_nullable`, `not_null`, and `nullable`.

qzyu999:
> TY for the updates and apologies for the response time, I left a couple more comments.
> 
> There's a failed CI documentation check run as well. I had a quick look but do not immediately know how the doc changes could have caused that.

Hi @leekeiabstraction, I've addressed your replies above. Regarding the CI documentation error, I think it maybe is related to a recent change in https://github.com/facebook/docusaurus/issues/11923. However, the latest build seems to be working.

fresh-borzoni (commented):
@qzyu999 Ty for the changes, looked through, PTAL

docusaurus was fixed here: https://github.com/apache/fluss-rust/pull/486
The column_writer.rs changes are good, both fixes are correct.                                                                                                                                                                                  
                                         
The datatype.rs additions aren't needed though. The only call site is the test, which can use IntType::with_nullable(false) that already exists. Adding public API surface just to make one test read slightly nicer isn't worth the maintenance cost, especially since as_non_nullable() already covers the pattern. 

Let's drop those and keep this PR focused on the actual fix.

qzyu999:
Hi @fresh-borzoni, I've reverted the changes to `datatype.rs` in b5b0b34 and so the changes so the PR is now just focused on the core fix in `column_writer.rs`. I also reran the tests locally and everything is passing.

fresh-borzoni (approved):
+1

leekeiabstraction (approved):
+1 thank you for your contribution

PR #477, opened 2026-04-05
fix: add missing write_array impl to IcebergBinaryRowWriter

leekeiabstraction:
## Summary
- The `BinaryWriter` trait gained a `write_array` method (from the array data type support PR #433), but `IcebergBinaryRowWriter` was not updated, causing a build failure on main.
- Added the missing `write_array` implementation that panics with a clear message, since array values are unsupported as Iceberg bucket keys (consistent with `set_null_at` and the existing `create_value_writer` rejection).

## Test plan
- [x] `cargo build --workspace --all-targets` passes
- [x] `cargo test --workspace` — all 312 unit tests pass


leekeiabstraction:
@fresh-borzoni Appreciate if you can review this small PR to fix CI failure on `main`

fresh-borzoni (approved):
+1

PR #495, opened 2026-04-17
[python] make LogScanner poll methods async to prevent event loop blocking

### LINKED ISSUE #496: [python ] LogScanner poll API is sync and blocks the asyncio event loop
### Search before asking

### Description

LogScanner's poll(), poll_record_batch(), poll_arrow(), to_arrow() and to_pandas() are synchronous and internally do py.detach + block_on. Because they run on the event loop thread, concurrent future_into_py tasks can't make progress while a poll is in flight, which causes intermittent deadlocks when users mix scanning with other async operations. It's also inconsistent with the rest of the Python API, which is fully async.                                                          
                                                            
Convert these five methods to async via future_into_py. This is a breaking change - they return awaitables. All call sites, examples, docs and type stubs need updating. The private _async_poll helpers used by aiter become redundant once this lands and should be removed, and the function-scoped test fixtures added in #494 (a workaround for the blocking loop) can be reverted.

### Willingness to contribute

fresh-borzoni:
closes https://github.com/apache/fluss-rust/issues/496

Converts LogScanner's poll(), poll_record_batch(), poll_arrow(), to_arrow() and to_pandas() from sync (py.detach + block_on) to async via future_into_py, so scanning no longer blocks the asyncio event loop.                                                                                                            
                                                            
With the public API now async, the private _async_poll and _async_poll_batches helpers are redundant and have been removed. Test fixtures back to session level and don't use global connection caching.
                                                                                                                                                                                                                                                     
Per-bucket stopping-offset tracking in to_arrow/to_pandas, timeout validation, partition handling and error propagation are preserved.

### DIFF
--- a/bindings/python/example/example.py
+++ b/bindings/python/example/example.py
@@ -278,7 +278,7 @@ async def main():
 
         # Try to get as PyArrow Table
         try:
-            pa_table_result = batch_scanner.to_arrow()
+            pa_table_result = await batch_scanner.to_arrow()
             print(f"\nAs PyArrow Table: {pa_table_result}")
         except Exception as e:
             print(f"Could not convert to PyArrow: {e}")
@@ -289,7 +289,7 @@ async def main():
 
         # Try to get as Pandas DataFrame
         try:
-            df_result = batch_scanner2.to_pandas()
+            df_result = await batch_scanner2.to_pandas()
             print(f"\nAs Pandas DataFrame:\n{df_result}")
         except Exception as e:
             print(f"Could not convert to Pandas: {e}")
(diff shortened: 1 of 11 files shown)


fresh-borzoni:

leekeiabstraction (approved):
LGTM, TY for the PR!

leekeiabstraction:
Ah, seemed like the CI has failed on main after this merge.

https://github.com/apache/fluss-rust/actions/runs/25278873151/job/74112896178

Appreciate if you can have a look @fresh-borzoni

fresh-borzoni:
@leekeiabstraction already taking a look, np

fresh-borzoni:
addressed: https://github.com/apache/fluss-rust/pull/527

PR #500, opened 2026-04-18
[rust] prefix lookup support

### LINKED ISSUE #499: [rust] prefix lookup support
### Search before asking

### Description

 Adds prefix lookup to the Rust SDK, mirroring the Java client's `PrefixKeyLookuper`.

### Willingness to contribute

fresh-borzoni:
closes https://github.com/apache/fluss-rust/issues/499

Adds prefix lookup to the Rust SDK. Users can now scan every row whose primary key starts with a given prefix via `table.new_lookup().lookup_by(cols).create_lookuper()`, matching the Java client's `PrefixKeyLookuper` API and validation.

`LookupQuery` is split into a typed enum of `PrimaryLookupQuery` and `PrefixLookupQuery` -  type-mismatched completion is now a compile error instead of a silent oneshot drop. Shared retry and error logic moved behind two module-private traits so primary and prefix paths share code without `dyn` dispatch.

 Two performance wins came along. Lookup keys stay as `Bytes` end-to-end (prost `bytes=` on request protos), so no `.to_vec()` on the hot path. And the 100ms poll in `LookupQueue::drain` is replaced by a `watch`-based wake off `Metadata` - re-enqueued lookups stranded on an unknown leader retry within microseconds of a metadata refresh instead of up to 100ms later. Java still has that `Thread.sleep(100)`.


fresh-borzoni:

fresh-borzoni:
fixed fmt

crates/fluss/src/client/metadata.rs
```diff
@@ -26,24 +26,36 @@ use parking_lot::RwLock;
...
         })
     }
 
+    pub fn subscribe_cluster_changes(&self) -> watch::Receiver<u64> {
+        self.cluster_version_tx.subscribe()
+    }
+
+    fn notify_cluster_changed(&self) {
+        let v = *self.cluster_version_tx.borrow();
+        let _ = self.cluster_version_tx.send(v.wrapping_add(1));
```
leekeiabstraction: Suggest the following to avoid potential race condition

```rust
self.cluster_version_tx.send_modify(|v| *v = v.wrapping_add(1));
```

crates/fluss/src/client/lookup/lookup_query.rs
```diff
@@ -15,77 +15,177 @@
...
-
+use crate::error::{Error, Result};
 use crate::metadata::{TableBucket, TablePath};
 use bytes::Bytes;
 use tokio::sync::oneshot;
 
-/// Represents a single lookup query that will be batched and sent to the server.
-pub struct LookupQuery {
-    /// The table path for this lookup
+struct QueryShared {
```
leekeiabstraction: Some amount of code duplication in `PrimaryLookupQuery` and `PrefixLookupQuery`. I wonder if we can do the following:

```rust
  pub struct LookupQuery<T> {
      table_path: TablePath,
      table_bucket: TableBucket,
      key: Bytes,
      retries: i32,                                                                                                                   
      result_tx: Option<oneshot::Sender<Result<T>>>,
  }                                                                               
  pub type PrimaryLookupQuery = LookupQuery<Option<Vec<u8>>>;
  pub type PrefixLookupQuery  = LookupQuery<Vec<Vec<u8>>>;

... shared impl
```

crates/fluss/src/client/lookup/lookup_queue.rs
```diff
@@ -78,32 +80,41 @@ impl LookupQueue {
...
-                    lookups.push(lookup);
-                    // Try to drain more without waiting
-                    while lookups.len() < self.max_batch_size {
-                        match self.lookup_rx.try_recv() {
-                            Ok(lookup) => lookups.push(lookup),
-                            Err(_) => break,
+            let sleep = tokio::time::sleep(remaining);
+            tokio::select! {
+                biased;
+
```
leekeiabstraction: nit: extra space here

crates/fluss/src/client/lookup/lookup_sender.rs
```diff
@@ -35,52 +30,100 @@ use std::collections::{HashMap, HashSet};
...
     running: AtomicBool,
-    /// Whether to force close (abandon pending lookups)
     force_close: AtomicBool,
-    /// Shutdown signal receiver
     shutdown_rx: watch::Receiver<bool>,
 }
 
-/// A batch of lookups going to the same table bucket.
-struct LookupBatch {
+struct PrimaryLookupBatch {
```
leekeiabstraction: Similarly here, can we use generics to reduce code duplication?

leekeiabstraction (commented):
TY for the PR, left some comments.

fresh-borzoni:
@leekeiabstraction Ty for the review, addressed comments, PTAL 🙏

crates/fluss/src/client/table/lookup.rs
```diff
@@ -314,6 +315,209 @@ impl Lookuper {
...
+    if bucket_keys.is_empty() {
+        return Err(Error::IllegalArgument {
+            message: format!(
+                "Can not perform prefix lookup on table '{}', because it has no bucket keys.",
+                table_info.get_table_path()
+            ),
+        });
+    }
+
+    for (i, bucket_key) in bucket_keys.iter().enumerate() {
```
leekeiabstraction: This can be simplified to 
```rust
if !physical_primary_keys.starts_with(bucket_keys) {
```

leekeiabstraction (commented):
TY for the update. Added further comment. Should we have doc, example and IT tests included in the PR as well?

crates/fluss/src/client/table/lookup.rs
fresh-borzoni: Good catch, ty!

fresh-borzoni:
@leekeiabstraction added docs and examples, integration tests we already have. PTAL 🙏

crates/fluss/src/client/lookup/lookup_sender.rs
```diff
@@ -233,38 +299,48 @@ impl LookupSender {
...
                 );
             }
         }
 
-        // If no lookup batches, sleep a bit to avoid busy loop. This case will happen when there is
-        // no leader for all the lookup request in queue.
-        if lookup_batches.is_empty() && !self.queue.has_undrained() {
-            // TODO: May use wait/notify mechanism to avoid active sleep, and use a dynamic sleep time based on the request waited time.
-            tokio::time::sleep(Duration::from_millis(100)).await;
+        if group.is_empty() && !self.queue.has_undrained() {
```
leekeiabstraction: In the case where leader is not available, group_by_leader re-enqueue immediately. Will `!self.queue.has_undrained()` cause the guard to not activate and therefore the intended 100ms sleep not fired?

crates/fluss/src/client/table/lookup.rs
```diff
@@ -314,6 +315,195 @@ impl Lookuper {
...
+    metadata: Arc<Metadata>,
+    lookup_client: Arc<LookupClient>,
+    bucketing_function: Box<dyn BucketingFunction>,
+    prefix_key_encoder: Box<dyn KeyEncoder>,
+    partition_getter: Option<PartitionGetter>,
+    num_buckets: i32,
+}
+
+impl PrefixKeyLookuper {
+    pub async fn lookup(&mut self, row: &dyn InternalRow) -> Result<LookupResult> {
```
leekeiabstraction, line 467: Java side does a metadata update, should we replicate the behaviour here?

crates/fluss/src/client/table/lookup.rs
leekeiabstraction, line 101: Do we need to address this todo?

leekeiabstraction: We can update/remove this.

leekeiabstraction (commented):
TY for the update. Left a few more smaller comments

crates/fluss/src/client/table/lookup.rs
fresh-borzoni: yes, good catch, also added for primary key path

crates/fluss/src/client/lookup/lookup_sender.rs
fresh-borzoni: Hmm, restructured code to handle this better

crates/fluss/src/client/table/lookup.rs
fresh-borzoni: I'd prefer to defer this and handle separately, this isn't trivial, so better be left for follow-up, also we might need to check other places in KV for schema evolution specifically.

fresh-borzoni:
@leekeiabstraction Ty for good review, addressed comments, PTAL 🙏

crates/fluss/src/client/table/lookup.rs
fresh-borzoni: added an issue to address this separately: https://github.com/apache/fluss-rust/issues/508

leekeiabstraction:
TY for the PR!

leekeiabstraction (approved):
+1

PR #501, opened 2026-04-18
[c++] Add c++ support for Array data type

### LINKED ISSUE #468: [c++] Add Array data support for c++ binding
### Search before asking

### Description

Follow-up to #433  which added end-to-end ARRAY column support in Rust core.

The Python bindings need to be updated to expose array support to users; the C++ bindings have partial passthrough and need full user-facing support.

### Willingness to contribute

charlesdong1991:
### Purpose

Linked issue: close #468 

### Brief change log
- Adds ARRAY data type support to the C++ bindings and cover full write and read paths (by creating ArrayWriter and ArrayView, also expose in public api)
- Support nested arrays on both write and read 
- Extend GenericRow, RowView and LookupResult with array-typed getters and setters

### Tests

All tests are passed

### DIFF
--- a/bindings/cpp/include/fluss.hpp
+++ b/bindings/cpp/include/fluss.hpp
@@ -620,8 +620,8 @@ class ArrayView {
     ArrayView(ArrayView&& other) noexcept;
     ArrayView& operator=(ArrayView&& other) noexcept;
 
-    size_t Size() const;
-    TypeId ElementType() const;
+    size_t Size() const noexcept;
+    TypeId ElementType() const noexcept;
     bool IsNull(size_t element) const;
 
     bool GetBool(size_t element) const;
@@ -664,6 +664,7 @@ class ArrayWriter {
     ArrayWriter& operator=(ArrayWriter&& other) noexcept;
 
     bool Available() const;
+    size_t Size() const noexcept;
 
     void SetNull(size_t idx);
     void SetBool(size_t idx, bool v);
--- a/bindings/cpp/include/fluss.hpp
+++ b/bindings/cpp/include/fluss.hpp
@@ -50,6 +50,8 @@ struct Lookuper;
 struct ScanResultInner;
 struct GenericRowInner;
 struct LookupResultInner;
+struct ArrayWriterInner;
+struct ArrayViewInner;
 }  // namespace ffi
 
 /// Named constants for Fluss API error codes.
@@ -276,6 +278,7 @@ enum class TypeId {
     Decimal = 14,
     Char = 15,
     Binary = 16,
+    Array = 17,
 };
 
 class DataType {
@@ -305,15 +308,29 @@ class DataType {
     }
     static DataType Char(int32_t length) { return DataType(TypeId::Char, length, 0); }
     static DataType Binary(int32_t length) { return DataType(TypeId::Binary, length, 0); }
+    /// Constructs an `ARRAY<element>` type. The element DataType (possibly
+    /// itself an array) is deep-copied into a shared owning handle so that
+    /// copies of the outer DataType remain cheap while the element lives
+    /// as long as any reference exists.
+    static DataType Array(DataType element) {
+        DataType dt(TypeId::Array, 0, 0);
+        dt.element_type_ = std::make_shared<DataType>(std::move(element));
+        return dt;
+    }
 
     TypeId id() const { return id_; }
     int32_t precision() const { return precision_; }
     int32_t scale() const { return scale_; }
+    /// Returns the element type of an ARRAY. Returns `nullptr` for non-array
+    /// types. The returned pointer is valid as long as this DataType (or a
+    /// copy holding the same shared element) is alive.
+    const DataType* element_type() const { return element_type_.get(); }
 
    private:
     TypeId id_;
     int32_t precision_{0};
     int32_t scale_{0};
+    std::shared_ptr<DataType> element_type_;
 };
 
 constexpr int64_t EARLIEST_OFFSET = -2;
@@ -493,6 +510,12 @@ inline size_t ResolveColumn(const ColumnMap& map, const std::string& name) {
     return it->second.index;
 }
 
+// Forward declaration so NamedGetters can declare GetArrayView(...) even
+// though the concrete class is defined further down.
+}  // namespace detail
+class ArrayView;
+namespace detail {
+
 /// CRTP mixin that adds name-based getters to any class with index-based getters.
 /// Derived must provide: `size_t Resolve(const std::string&) const`
 /// and all the index-based getters (IsNull(idx), GetBool(idx), etc.).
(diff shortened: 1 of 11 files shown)


charlesdong1991:
@fresh-borzoni @leekeiabstraction appreciate it a lot if you can take a look! 🙏 thanks!

bindings/cpp/src/table.cpp
```diff
@@ -85,6 +85,219 @@ int Date::Day() const {
...
+    inner_->aw_set_decimal_str(idx, value);
+}
+
+void ArrayWriter::SetArray(size_t idx, ArrayWriter&& nested) {
+    CHECK_AW("ArrayWriter");
+    if (!nested.inner_) {
+        throw std::logic_error("ArrayWriter::SetArray: nested writer not available");
+    }
+    inner_->aw_set_array(idx, *nested.inner_);
+}
```
leekeiabstraction, line 191: Not familiar with C++, but it seems like the convention of `left::method(right&& arg)` is for left to take ownership.

https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#rf-consume

Should we call `Destroy()` to ensure no memory leak?

bindings/cpp/src/table.cpp
```diff
@@ -191,6 +404,14 @@ void GenericRow::SetDecimal(size_t idx, const std::string& value) {
     inner_->gr_set_decimal_str(idx, value);
 }
 
+void GenericRow::SetArray(size_t idx, ArrayWriter&& writer) {
```
leekeiabstraction, line 415: Similarly here on calling `Destroy()`

leekeiabstraction (commented):
TY for the PR. Left early comments.

bindings/cpp/src/table.cpp
charlesdong1991: i dont think it will cause memory leak since when nested goes out of scope at caller, ArrayWriter will run and call destroy, and will free `Box` from rust, so cleanup will happen.

But with another thought, i think it is still worth adding here explicitly, since it is aligned with c++ convention, let me update

bindings/cpp/src/table.cpp
charlesdong1991: the same as above

bindings/cpp/include/fluss.hpp
```diff
@@ -305,15 +308,29 @@ class DataType {
     }
     static DataType Char(int32_t length) { return DataType(TypeId::Char, length, 0); }
     static DataType Binary(int32_t length) { return DataType(TypeId::Binary, length, 0); }
+    /// Constructs an `ARRAY<element>` type. The element DataType (possibly
+    /// itself an array) is deep-copied into a shared owning handle so that
+    /// copies of the outer DataType remain cheap while the element lives
+    /// as long as any reference exists.
+    static DataType Array(DataType element) {
```
fresh-borzoni, line 315: Should we add element nullability?

Rust's ArrayType::with_nullable(false, e) carries it, by dropping here we silently lose ARRAY<X NOT NULL> on schema round-trip. Probably the same applies to python, do you mind to check?

bindings/cpp/include/fluss.hpp
```diff
@@ -305,15 +308,29 @@ class DataType {
...
+    /// Returns the element type of an ARRAY. Returns `nullptr` for non-array
+    /// types. The returned pointer is valid as long as this DataType (or a
+    /// copy holding the same shared element) is alive.
+    const DataType* element_type() const { return element_type_.get(); }
 
    private:
     TypeId id_;
     int32_t precision_{0};
     int32_t scale_{0};
+    std::shared_ptr<DataType> element_type_;
```
fresh-borzoni, line 333: also is_nullable field needed here

bindings/cpp/include/fluss.hpp
```diff
@@ -535,6 +603,90 @@ struct ScanData {
...
+class ArrayView {
+   public:
+    ~ArrayView() noexcept;
+
+    ArrayView(const ArrayView&) = delete;
+    ArrayView& operator=(const ArrayView&) = delete;
+    ArrayView(ArrayView&& other) noexcept;
+    ArrayView& operator=(ArrayView&& other) noexcept;
+
+    size_t Size() const;
```
fresh-borzoni: noexcept for Size and ElementType?

bindings/cpp/src/ffi_converter.hpp
```diff
@@ -20,13 +20,72 @@
...
+/// scalar. A non-array input produces a zero-initialised value (nesting == 0).
+///
+/// Using a flat representation — rather than serialising a recursive
+/// `DataType` — keeps the cxx bridge contract small (four `i32`s inside
+/// `FfiColumn`) while preserving full schema fidelity across the FFI boundary
+/// when paired with rebuild_array_type().
+struct FlattenedArrayType {
+    int32_t nesting{0};
+    int32_t leaf_type{0};
+    int32_t leaf_precision{0};
```
fresh-borzoni, line 44: also thread nullability here

bindings/cpp/include/fluss.hpp
```diff
@@ -535,6 +603,90 @@ struct ScanData {
...
+};
+
+namespace detail {
+template <typename Derived>
+inline ArrayView NamedGetters<Derived>::GetArrayView(const std::string& n) const {
+    return Self().GetArrayView(Self().Resolve(n));
+}
+}  // namespace detail
+
+class ArrayWriter {
```
fresh-borzoni, line 656: we have num elements as well if I recall correctly

bindings/cpp/test/test_log_table.cpp
```diff
@@ -829,3 +829,436 @@ TEST_F(LogTableTest, PartitionedTableAppendScan) {
 
     ASSERT_OK(adm.DropTable(table_path, false));
 }
+
+// ============================================================================
+// Array data type tests
```
fresh-borzoni, line 836: I am thinking about these edge cases as well:

- Long string in array (≥8 bytes for variable-length encoding path)                                                                                                                                                                              
- ARRAY<DECIMAL(20+, 5)> for non-compact decimal                                                                                                                                                                                                 
- ARRAY<TIMESTAMP(9)> for non-compact timestamp                                                                                                                                                                                                  
- Float NaN for  ±Infinity round-trip                                                                                                                                                                                                             
- ARRAY<BINARY(N)> for fixed-length                                                                                                                                                                                        
- SetInt32(1000) on TINYINT must throw

fresh-borzoni (commented):
@charlesdong1991 Ty, looks good in general, added minor comments, PTAL

bindings/cpp/include/fluss.hpp
charlesdong1991: Good observation, with another look, it seems nullability is actually not in place for every data type, not limited to Array. The same with python.

I suggest create a separate issue to fix nullability instead, and focus on supporting array type in this PR. 

WDYT? If you agree, i can create 2 issues for python and c++ @fresh-borzoni

bindings/cpp/include/fluss.hpp
charlesdong1991: same above

bindings/cpp/include/fluss.hpp
charlesdong1991: done, i also expose Size to arraywriter for symmetry with arrayview

bindings/cpp/test/test_log_table.cpp
charlesdong1991: nice!

bindings/cpp/include/fluss.hpp
charlesdong1991: changed!

bindings/cpp/include/fluss.hpp
fresh-borzoni: Sure, it works

charlesdong1991: Created 2 issues

fresh-borzoni (approved):
@charlesdong1991 Ty for the great work! LGTM 👍

luoyuxia (approved):

PR #517, opened 2026-05-01
[rust] lookup decoding schema id 

### LINKED ISSUE #508: [rust] Lookup result decodes every row with the latest schema, ignoring per-row schema id
### Search before asking

### Description

Every row stored in Fluss carries a 2-byte schema id prefix. The client's LookupResult currently ignores that id and decodes every row with self.row_type (the latest schema of the table). If a row was written under an older schema version, decoding it under the latest RowType produces wrong field values.
                                                                                                                                                                                                                                                     This affects both Lookuper::lookup (single row) and PrefixKeyLookuper::lookup (multiple rows, any of which could be from any write time).

### Willingness to contribute

fresh-borzoni:
## Summary

closes #508

Lookup rows now decode under the schema id they were written with, then project to the table's current schema. Index mapping matches by `column id` (stable across `ALTER TABLE … RENAME`).                                                                                                                                                                                              
                                                                                                                                                                                                                                                     Side fix: `Column.id` is now preserved through `Schema` JSON serde as it was silently dropped on the read path.


Cargo.lock
```diff
@@ -746,6 +746,15 @@ dependencies = [
  "rustc_version",
```
fresh-borzoni, line 746: I will rebase once we sync Cargo.lock

fresh-borzoni:
@leekeiabstraction @charlesdong1991 PTAL 🙏

crates/fluss/src/client/table/lookup.rs
```diff
@@ -16,118 +16,230 @@
...
         }
     }
 
-    /// Extracts the row payload by stripping the schema id prefix.
-    fn extract_payload(bytes: &[u8]) -> Result<&[u8]> {
-        bytes
-            .get(SCHEMA_ID_LENGTH..)
-            .ok_or_else(|| Error::RowConvertError {
+    fn read_schema_id(bytes: &[u8]) -> Result<i16> {
+        if bytes.len() < SCHEMA_ID_LENGTH {
```
charlesdong1991, line 110: i am curious in which scenario will payload only contain schema id?

crates/fluss/src/row/projected_row.rs
```diff
@@ -0,0 +1,276 @@
...
+            .get(pos)
+            .copied()
+            .ok_or_else(|| IllegalArgument {
+                message: format!(
+                    "position {pos} out of bounds (projected row has {} fields)",
+                    self.index_mapping.len()
+                ),
+            })?;
+        if mapped == UNEXIST_MAPPING {
+            return Err(IllegalArgument {
```
charlesdong1991, line 54: so users who call get_rows might catch this error?

crates/fluss/src/row/lookup_row.rs
```diff
@@ -0,0 +1,122 @@
...
+        delegate!(self, get_binary, pos, length)
+    }
+    fn get_bytes(&self, pos: usize) -> Result<&[u8]> {
+        delegate!(self, get_bytes, pos)
+    }
+    fn get_array(&self, pos: usize) -> Result<FlussArray> {
+        delegate!(self, get_array, pos)
+    }
+    fn as_encoded_bytes(&self, write_format: WriteFormat) -> Option<&[u8]> {
+        delegate!(self, as_encoded_bytes, write_format)
```
charlesdong1991, line 120: just for my own understanding, in this case, payload byte will have no schema id, right?

charlesdong1991 (commented):
very nice PR!! just some nit comments!

crates/fluss/src/row/lookup_row.rs
fresh-borzoni: yes, it mirrors Java's `decode(segment, SCHEMA_ID_LENGTH, size - SCHEMA_ID_LENGTH)`

crates/fluss/src/row/projected_row.rs
fresh-borzoni: Yes, but only if they skip is_null_at. Java throws some kind of OOB errors, I preferred the exact exception here

crates/fluss/src/client/table/lookup.rs
fresh-borzoni: Pure defensive bounds-check. It's mostly against truncated payload, so if there is less than 2 bytes - fail here while reading the schema, if there is empty body - then decoder will fail with more specific error.

Practically, I don't think it's possible to have correct payload with only schema id, but I thought this way :)

fresh-borzoni:
@charlesdong1991 Ty for the review, answered comments, PTAL 🙏

crates/fluss/src/row/projected_row.rs
charlesdong1991: gotcha, thanks!

crates/fluss/src/client/admin.rs
```diff
@@ -111,6 +111,27 @@ impl FlussAdmin {
         Ok(())
     }
 
+    /// Fetch the schema for `table_path` at the given `schema_id`. Pass
+    /// `None` to request the latest.
+    pub async fn get_table_schema(
```
charlesdong1991, line 116: i think we should update api-reference on this

crates/fluss/src/metadata/table.rs
```diff
@@ -152,6 +169,31 @@ impl Schema {
...
 }
 
+/// A schema together with its server-assigned version id.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct SchemaInfo {
+    schema: Schema,
+    schema_id: i32,
+}
+
+impl SchemaInfo {
```
charlesdong1991, line 179: should we add this too to api-reference?

charlesdong1991 (approved):
the implementation looks good to me! great job!! @fresh-borzoni 

Only a nitpick: i think we could update api-references a bit, not blocking and can be separate pr too.

crates/fluss/src/client/admin.rs
leekeiabstraction, line 116: I think we are currently blocked on creating end to end test to fully test this PR's change, specifically blocked by the lack of alter table support on rust side. I've raised #526 to signify this gap.

crates/fluss/src/client/table/lookup.rs
```diff
@@ -16,118 +16,230 @@
...
+            return Some(Arc::clone(&self.target_decoder));
+        }
+        self.others.read().get(&schema_id).cloned()
+    }
+}
+
+/// Rows returned from a lookup. Primary-key lookups produce at most one
+/// row; prefix-key lookups may produce many. Rows written under older
+/// schemas are decoded with their original schema and projected to the
+/// table's current schema before the caller sees them.
```
leekeiabstraction: When we say current schema, do we mean at the time when lookuper was created or always latest? From what I can see, rows are projected to the table's schema at the point in time when lookuper was created.

leekeiabstraction (commented):
TY for the PR, left a couple of comments.

crates/fluss/src/client/admin.rs
fresh-borzoni: Yes, that’s true. The main goal of this PR is to make sure Rust clients can safely read Fluss data that may have been written before a schema change.

I agree that full alter table support on the Rust side is needed, and thanks for opening #526 to track that gap. I’d still prefer to keep that as a separate scope: first land the schema-aware lookup/decoding support, then add alter table support, and finally add the full end-to-end test once the Rust client can actually perform the schema change itself. That feels like a proper sequence, wdyt?

crates/fluss/src/client/table/lookup.rs
fresh-borzoni: you're right, the wording is misleading. The schema is captured when the Lookuper is constructed, and stays pinned for the lifetime of that Lookuper - schema evolutions that land afterwards aren't picked up and the caller has to create a new Lookuper to see them.
 
 I'll change the wording to sound less confusing, ty!

fresh-borzoni: updated

fresh-borzoni:
@leekeiabstraction Ty for the review, addressed all comments, PTAL 🙏

crates/fluss/src/client/admin.rs
fresh-borzoni: Good catch, ty!

crates/fluss/src/client/table/lookup.rs
```diff
@@ -16,118 +16,232 @@
...
-/// The result of a lookup operation.
-///
-/// Contains the rows returned from a lookup. For primary-key lookups,
-/// this will contain at most one row. For prefix-key lookups, it may
-/// contain multiple rows.
+/// Per-Lookuper decoder cache. The target-schema decoder is held
+/// directly so the dominant decode path is a single field access; older
+/// schemas are populated lazily on first observation.
+struct DecoderCache {
+    target_id: i16,
```
leekeiabstraction, line 42: We have a few places where schema id is passed either as i32 or i16. I wonder if consolidating into a single schema id type is cleaner. That way we can define `from_wire/to_wire/to_api` on the type where wire / API integer value is actually. This will make passing of the arg easier from calling site. What do you think?

crates/fluss/src/metadata/datatype.rs
```diff
@@ -93,6 +93,60 @@ impl DataType {
             DataType::Bytes(v) => DataType::Bytes(v.as_non_nullable()),
         }
     }
+
+    /// Structural equality ignoring the outermost nullability flag at
+    /// every level. Equivalent to comparing `as_non_nullable()` on both
+    /// sides but without the recursive clone.
+    pub fn eq_ignore_nullable(&self, other: &DataType) -> bool {
```
leekeiabstraction: Should we consider `pub(crate)`? We can consider promoting to `pub` if there's signal that this is useful for end users.

crates/fluss/src/metadata/schema_util.rs
```diff
@@ -0,0 +1,204 @@
...
+use std::collections::{HashMap, HashSet};
+
+/// Sentinel for an expected column that does not exist in the origin
+/// schema. Used by [`index_mapping`] and [`crate::row::ProjectedRow`].
+pub const UNEXIST_MAPPING: i32 = -1;
+
+/// For each column in `expected_schema`, return the index of the column
+/// with the same id in `origin_schema`, or [`UNEXIST_MAPPING`] if absent.
+/// Matching by id keeps mappings stable across `ALTER TABLE … RENAME`.
+pub fn index_mapping(origin_schema: &Schema, expected_schema: &Schema) -> Result<Vec<i32>> {
```
leekeiabstraction: Should we consider `pub(crate)`? We can consider promoting to `pub` if there's signal that this is useful for end users.

leekeiabstraction (commented):
TY for addressing, left further comments.

crates/fluss/src/client/table/lookup.rs
fresh-borzoni: Yes, this is a good point, I'll file an issue to cleanup this, though java has the same mess tbh, so we might want to reason and relate this back to java 🤔

fresh-borzoni: https://github.com/apache/fluss-rust/issues/528 - ✅

fresh-borzoni:
@leekeiabstraction Ty, addressed 👍

leekeiabstraction (commented):
TY for the PR. LGTM

PR #407, opened 2026-03-01
feat: [metrics] Create metrics framework and add connection metrics

charlesdong1991:
### Purpose

Linked issue: close #390 

### Brief change log

This PR adds a metrics framework and connection level metrics.

NOTE: 
There is an parity from Java, which is the scopeguard for in-flight on cancellation, because Java callback model doesn't have equivalent to tokio future cancellation. And if a tokio future is dropped mid-execution, it would skip decrement and cause gauge drift. So i think it's important to add "scopeguard" to avoid that.

And I can have writer and scanner metrics as follow-up PRs.

### Tests

All tests are passed locally.

PR #518, opened 2026-05-01
chore: regenerate Cargo.lock for declared metrics deps

fresh-borzoni:
After https://github.com/apache/fluss-rust/pull/407 we have Cargo.lock not synced, so this PR fixes this issue


fresh-borzoni:
@leekeiabstraction PTAL 🙏

PR #525, opened 2026-05-03
[c++] Add nullability support for data type in c++

### LINKED ISSUE #510: [c++] Add nullability support to DataType in C++ binding
### Search before asking

### Description

The C++ `DataType` class does not carry a `nullable` flag. In Rust, every `DataType` variant stores `nullable: bool`, and composite types like `ArrayType` carry nullability at each nesting level. The C++ binding silently drops this information.

See comment in #501 

### Willingness to contribute

charlesdong1991:
### Purpose

Preserve nullability when constructing Fluss schemas through the C++ binding. Previously, the C++ `DataType` class had no nullable flag, so all types defaulted to nullable regardless of what the user specified.

Linked issue: close #510 

### Tests

All tests have passed locally

### Documentation

Docs are updated accordingly.


bindings/cpp/src/types.rs
```diff
@@ -354,9 +406,31 @@ pub fn element_type_from_ffi(
...
+            data_type: leaf_dt,
+            precision,
+            scale,
+            element_data_type: 0,
+            element_precision: 0,
+            element_scale: 0,
+            array_nesting: 0,
+            array_nullability: Vec::new(),
+            nullable: true,
+            element_nullable: true,
```
leekeiabstraction: Can you elaborate why and the impact of hardcoding nullability here? Also other places where element nullability is hardcoded to true.

leekeiabstraction (commented):
TY for the PR, left a question.

bindings/cpp/src/types.rs
charlesdong1991: hardcoding here is on purpose actually, since the spec struct has 2 nullability concept, where nullable is the nullability of this constructed type where element_nullable is for the leaf element of an array.

and function like `build_array_type_from_leaf` recursively calls to construct leaf which is scaler, so element_nullable is unused.

ArrayWriter it only needs the type for encoding i believe, and putting default as true is aligned with Fluss default, which means all types are nullable unless explicitly marked.

I think probably best to add some comments in case of questions in future, WDYT? @leekeiabstraction

charlesdong1991:
thanks for the question. i replied, let me know if further question @leekeiabstraction 

if find time, would be great to give a look @fresh-borzoni @leekeiabstraction 🙏

bindings/cpp/src/types.rs
```diff
@@ -41,16 +41,33 @@ pub const DATA_TYPE_CHAR: i32 = 15;
 pub const DATA_TYPE_BINARY: i32 = 16;
 pub const DATA_TYPE_ARRAY: i32 = 17;
 
+struct FfiDataTypeSpec {
```
fresh-borzoni: Not sure this struct is the right shape. One struct lumps two jobs (build scalar / build array). Most call sites zero out the majority of the fields. Shall we split into separate Scalar/Array variants?

bindings/cpp/src/lib.rs
```diff
@@ -85,13 +85,16 @@ mod ffi {
...
+        nullable: bool,
         comment: String,
         precision: i32,
         scale: i32,
         array_nesting: i32,
+        array_nullability: Vec<u8>,
         element_data_type: i32,
         element_precision: i32,
         element_scale: i32,
+        element_nullable: bool,
```
fresh-borzoni: If we split FfiDataTypeSpec to Scalar/Array, this field can go away too = just make array_nullability one entry longer (last = leaf)

bindings/cpp/src/ffi_converter.hpp
```diff
@@ -72,16 +77,30 @@ inline FlattenedArrayType flatten_array_type(const DataType& data_type) {
...
-    for (int32_t i = 0; i < flat.nesting; ++i) {
-        dt = DataType::Array(std::move(dt));
+inline DataType rebuild_array_type(const FlattenedArrayType& flat, bool outer_nullable = true) {
+    DataType dt(static_cast<TypeId>(flat.leaf_type), flat.leaf_precision, flat.leaf_scale,
+                flat.leaf_nullable);
+    for (int32_t i = flat.nesting - 1; i >= 0; --i) {
+        bool nullable = true;
+        if (static_cast<size_t>(i) < flat.array_nullability.size()) {
+            nullable = flat.array_nullability[static_cast<size_t>(i)] != 0;
+        } else if (i == 0) {
```
fresh-borzoni: but do we have the case for this?

bindings/cpp/src/types.rs
```diff
@@ -41,16 +41,33 @@ pub const DATA_TYPE_CHAR: i32 = 15;
...
+    data_type: i32,
+    precision: u32,
+    scale: u32,
+    element_data_type: i32,
+    element_precision: u32,
+    element_scale: u32,
+    array_nesting: u32,
+    array_nullability: Vec<u8>,
+    nullable: bool,
+    element_nullable: bool,
```
fresh-borzoni: is it only for backward compatibility, mb we can drop it?

fresh-borzoni (commented):
@charlesdong1991 Ty for the PR, overall LGTM
Left some comments, PTAL

bindings/cpp/src/types.rs
leekeiabstraction: > I think probably best to add some comments in case of questions in future

Sounds good to me.

bindings/cpp/src/types.rs
charlesdong1991: yes, it's for backward compatibility, let me check this drop option! 👍

bindings/cpp/src/lib.rs
charlesdong1991: yeah, i think it's a nice idea to split as this is added a bit "ugly" for compatibility, i think splitting will be cleaner!

bindings/cpp/src/ffi_converter.hpp
charlesdong1991: this shouldn't be the case anymore after splitting

charlesdong1991:
thanks for great suggestion @fresh-borzoni i split the array from scaler, PTAL!

fresh-borzoni (approved):
@charlesdong1991 Ty, LGTM 👍

bindings/cpp/src/lib.rs
fresh-borzoni, line 1: Nit: doc comment on structurally_compatible (lib.rs:3617-3618) says "C++ callers never control nullability explicitly" - no longer true after this PR. 
Worth rewording to reflect the actual reason we ignore nullability: the Rust-side element type is always reconstructed as nullable (encoding doesn't depend on it), so comparing the bit is meaningless

fresh-borzoni:
@leekeiabstraction Do you have any further comments on this PR?

bindings/cpp/src/lib.rs
charlesdong1991: Good catch! 👍 rephrased in [62832ae](https://github.com/apache/fluss-rust/pull/525/commits/62832ae90e96a0812dfb47b85acec0b3106ec7ae) @fresh-borzoni

fresh-borzoni:
@charlesdong1991 Ty, merging

PR #529, opened 2026-05-03
[chore] reduce workers to 2 with loadfile distribution

charlesdong1991:
By a quick look, i have some ideas of where test flake can happen, and i think we can reduce concurrent workers from 4 to 2 (given unbuntu-latest probably has 4 cores)

Plus, i think usually `loadfile` could be a better distribution strategy than the default in concurrent runs scenario since tests from the same file will go to same worker now.

### DIFF
--- a/.github/workflows/build_and_test_python.yml
+++ b/.github/workflows/build_and_test_python.yml
@@ -77,7 +77,7 @@ jobs:
 
       - name: Run tests (parallel)
         working-directory: bindings/python
-        run: uv run --no-sync pytest test/ -v -n auto
+        run: uv run --no-sync pytest test/ -v -n 2 --dist=loadfile
         env:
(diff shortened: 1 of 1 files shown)


charlesdong1991:
not sure how much it will help, but i think it can help a bit, maybe we can give it a try? WDYT? @leekeiabstraction @fresh-borzoni 

there are some other improvements can be made too to improve test stability, and can share later.

charlesdong1991:
otherwise, i can close this PR 😄

fresh-borzoni (approved):
@charlesdong1991 

Thank you, let's try this as a quick fix, if you can propose some improvements in tests - also much appreciated 👍

leekeiabstraction (approved):

leekeiabstraction:
TY for the quick workaround!

PR #530, opened 2026-05-05
feat: implement FlussMap support for compacted rows and binary encoding

### LINKED ISSUE #387: Map data type support in Rust
### Search before asking

### Description

Depends on #386 
Add end-to-end support for MAP columns. A Map is stored as two parallel BinaryArrays (keys + values) prefixed by the key array's byte size. This builds directly on the BinaryArray infrastructure from #386 

The Datum enum needs a Map variant, InternalRow needs get_map(). The BinaryMap format is simple,  look at Java's BinaryMap.java and MapSerializer.java. 

The compacted row writer/reader, value writer, and field getter all need Map support in the same locations as Array (the TODOs cover both). 

The key encoder should reject Map as a key type.

Tests should cover maps with various key/value type combinations, empty maps, and nullable values.

Java reference: BinaryMap.java, MapSerializer.java

### Willingness to contribute

qzyu999:
### Purpose

Linked issue: close #387 

This pull request introduces support for the MAP data type in the Fluss Rust client. It allows developers to define associative arrays (key-value pairs) in their table schemas and provides the underlying binary representation (FlussMap) necessary to process these fields, ensuring wire compatibility with the Java client's BinaryMap.

### Brief change log

*   **Metadata & Schema**: Added `MapType` to the `DataType` enum and implemented the `DataTypes::map()` factory. Updated `eq_ignore_nullable` to support recursive map comparison.
*   **Core Data Type**: Implemented `FlussMap` in `crates/fluss/src/row/binary_map.rs`. This provides a zero-copy, wire-compatible wrapper for the `[4B key_size] + [keys] + [values]` binary layout.
*   **Row Integration**: Extended the `InternalRow` trait with `get_map()` and implemented it across `GenericRow`, `CompactedRow`, `LookupRow`, `ProjectedRow`, and nested `FlussArray` accessors.
*   **Serialization/Deserialization**: 
    *   Implemented `write_map` in `BinaryWriter`, `CompactedRowWriter`, and `CompactedKeyWriter`.
    *   Updated `CompactedRowDeserializer` to handle `MAP` data type reconstruction.
    *   Added `FieldGetter` and `InnerFieldGetter` support for dynamically accessing map fields.
*   **Arrow Interoperability**:
    *   Implemented `ColumnarRow::get_map()` to convert Arrow `MapArray` structures into `FlussMap`.
    *   Implemented `Datum::Map::append_to` using Arrow's `MapBuilder` for seamless integration with the Arrow ecosystem.
    *   Updated `ColumnWriter` to support building Arrow `MapArray` outputs from Fluss rows.
*   **Bindings**: Updated Python, C++, and Elixir binding layers to support `Datum::Map` during serialization and row processing.
*   **Key Constraints**: Explicitly enforced the rejection of `MAP` types as primary or bucketing keys in the `CompactedKeyWriter`.

### Tests

Added unit tests covering the new functionality:
*   **Unit Tests**: Added comprehensive coverage in `binary_map.rs` for round-trips, empty maps, null values, and malformed buffer validation.
*   **Integration Tests**: Added 4 end-to-end integration tests in `compacted_row.rs` covering basic maps, nullable map columns, nested maps (`Map<String, Array<Int>>`), and empty maps.
*   **Interoperability Tests**: Added a `ColumnWriter` unit test to verify that `MAP` columns are correctly serialized into valid Arrow `MapArray` structures.
*   **Regression Tests**: Added `test_encode_map_rejected` to ensure `MAP` cannot be incorrectly used as a key type.

### API and Format

*   **API**: Yes. Users can now use `DataTypes::map()` for schemas and `row.get_map()` for data retrieval.
*   **Format**: Yes. This implements the memory/wire layout for map serialization, maintaining compatibility with the Java implementation.

### Documentation

Yes. This adds the `MAP` data type. User guides will need updating to reflect that `MAP` is now a supported type in the Rust client.

### DIFF
--- a/crates/fluss/src/record/arrow.rs
+++ b/crates/fluss/src/record/arrow.rs
@@ -994,6 +994,7 @@ impl LogRecordBatch {
         let record_batch = read_context.record_batch(data)?;
         let arrow_reader = ArrowReader::new_with_fluss_row_type(
             Arc::new(record_batch),
+            read_context.row_type.clone(),
             read_context.fluss_row_type().cloned(),
         );
         let log_record_iterator = LogRecordIterator::Arrow(ArrowLogRecordIterator {
@@ -1020,6 +1021,7 @@ impl LogRecordBatch {
             Some(record_batch) => {
                 let arrow_reader = ArrowReader::new_with_fluss_row_type(
                     Arc::new(record_batch),
+                    read_context.row_type.clone(),
                     read_context.fluss_row_type().cloned(),
                 );
                 LogRecordIterator::Arrow(ArrowLogRecordIterator {
@@ -1237,7 +1239,7 @@ pub fn to_arrow_type(fluss_type: &DataType) -> Result<ArrowDataType> {
                 Arc::new(Field::new(
                     "entries",
                     ArrowDataType::Struct(arrow_schema::Fields::from(entry_fields)),
-                    fluss_type.is_nullable(),
+                    false,
                 )),
                 false,
             )
@@ -1341,6 +1343,7 @@ pub(crate) fn from_arrow_type(arrow_type: &ArrowDataType) -> Result<DataType> {
 pub struct ReadContext {
     target_schema: SchemaRef,
     full_schema: SchemaRef,
+    row_type: Arc<RowType>,
     projection: Option<Projection>,
     is_from_remote: bool,
     fluss_row_type: Option<Arc<RowType>>,
@@ -1357,10 +1360,15 @@ struct Projection {
 }
 
 impl ReadContext {
-    pub fn new(arrow_schema: SchemaRef, is_from_remote: bool) -> ReadContext {
+    pub fn new(
+        arrow_schema: SchemaRef,
+        row_type: Arc<RowType>,
+        is_from_remote: bool,
+    ) -> ReadContext {
         ReadContext {
             target_schema: arrow_schema.clone(),
             full_schema: arrow_schema,
+            row_type,
             projection: None,
             is_from_remote,
             fluss_row_type: None,
@@ -1378,6 +1386,7 @@ impl ReadContext {
 
     pub fn with_projection_pushdown(
         arrow_schema: SchemaRef,
+        row_type: Arc<RowType>,
         projected_fields: Vec<usize>,
         is_from_remote: bool,
     ) -> Result<ReadContext> {
@@ -1442,6 +1451,7 @@ impl ReadContext {
         Ok(ReadContext {
             target_schema,
             full_schema: arrow_schema,
+            row_type,
             projection: Some(project),
             is_from_remote,
             fluss_row_type: None,
@@ -1635,15 +1645,17 @@ impl Iterator for ArrowLogRecordIterator {
 
 pub struct ArrowReader {
     record_batch: Arc<RecordBatch>,
+    row_type: Arc<RowType>,
     fluss_row_type: Option<Arc<RowType>>,
     row_column_indices: Arc<[usize]>,
 }
 
 impl ArrowReader {
-    pub fn new(record_batch: Arc<RecordBatch>) -> Self {
+    pub fn new(record_batch: Arc<RecordBatch>, row_type: Arc<RowType>) -> Self {
         let row_column_indices = arrow_row_column_indices(&record_batch);
         ArrowReader {
             record_batch,
+            row_type,
             fluss_row_type: None,
             row_column_indices,
         }
@@ -1651,6 +1663,7 @@ impl ArrowReader {
 
     pub fn new_with_fluss_row_type(
         record_batch: Arc<RecordBatch>,
+        row_type: Arc<RowType>,
         fluss_row_type: Option<Arc<RowType>>,
     ) -> Self {
         let row_column_indices = match &fluss_row_type {
@@ -1659,6 +1672,7 @@ impl ArrowReader {
         };
         ArrowReader {
             record_batch,
+            row_type,
             fluss_row_type,
             row_column_indices,
         }
@@ -1671,6 +1685,7 @@ impl ArrowReader {
     pub fn read(&self, row_id: usize) -> ColumnarRow {
         ColumnarRow::with_indices(
             self.record_batch.clone(),
+            self.row_type.clone(),
             row_id,
             self.fluss_row_type.clone(),
             self.row_column_indices.clone(),
@@ -1799,10 +1814,10 @@ mod tests {
                 Arc::new(Field::new(
                     "entries",
                     ArrowDataType::Struct(arrow_schema::Fields::from(vec![
-                        Field::new("key", ArrowDataType::Utf8, true),
+                        Field::new("key", ArrowDataType::Utf8, false),
                         Field::new("value", ArrowDataType::Int32, true),
                     ])),
-                    true,
+                    false,
                 )),
                 false,
             )
@@ -1819,6 +1834,21 @@ mod tests {
                 Field::new("f2", ArrowDataType::Utf8, true),
             ]))
         );
+    }
+
+    #[test]
+    fn test_arrow_map_schema_strictness() {
+        let map_type = DataTypes::map(DataTypes::string(), DataTypes::int());
+        let arrow_type = to_arrow_type(&map_type).unwrap();
+
+        if let ArrowDataType::Map(entries_field, _) = arrow_type {
+            assert!(
+                !entries_field.is_nullable(),
+                "Arrow Map 'entries' field must be strictly non-nullable"
+            );
+        } else {
+            panic!("Expected ArrowDataType::Map, got {:?}", arrow_type);
+        }
     }
 
     #[test]
@@ -1878,7 +1908,8 @@ mod tests {
             DataField::new("name", DataTypes::string(), None),
         ]);
         let schema = to_arrow_schema(&row_type).unwrap();
-        let result = ReadContext::with_projection_pushdown(schema, vec![0, 2], false);
+        let result =
+            ReadContext::with_projection_pushdown(schema, Arc::new(row_type), vec![0, 2], false);
 
         assert!(matches!(result, Err(IllegalArgument { .. })));
     }
--- a/crates/fluss/src/record/arrow.rs
+++ b/crates/fluss/src/record/arrow.rs
@@ -1261,6 +1261,16 @@ pub fn to_arrow_type(fluss_type: &DataType) -> Result<ArrowDataType> {
     })
 }
 
+/// Like `from_arrow_type`, but also reads the Field's nullability —
+/// Arrow stores it on the Field wrapper, not the leaf data type.
+pub(crate) fn from_arrow_field(field: &arrow_schema::Field) -> Result<DataType> {
+    let mut dt = from_arrow_type(field.data_type())?;
+    if !field.is_nullable() {
+        dt = dt.as_non_nullable();
+    }
+    Ok(dt)
+}
+
 /// Converts an Arrow data type back to a Fluss `DataType`.
 /// Used for reading array elements from Arrow ListArray back into Fluss types.
 pub(crate) fn from_arrow_type(arrow_type: &ArrowDataType) -> Result<DataType> {
@@ -1317,17 +1327,30 @@ pub(crate) fn from_arrow_type(arrow_type: &ArrowDataType) -> Result<DataType> {
                 DataTypes::timestamp_with_precision(precision)
             }
         }
-        ArrowDataType::List(field) => DataTypes::array(from_arrow_type(field.data_type())?),
+        ArrowDataType::List(field) => DataTypes::array(from_arrow_field(field)?),
+        ArrowDataType::Map(entries_field, _sorted) => {
+            let fields = match entries_field.data_type() {
+                ArrowDataType::Struct(f) => f,
+                other => {
+                    return Err(Error::IllegalArgument {
+                        message: format!("Map entries must be Struct, got {other:?}"),
+                    });
+                }
+            };
+            if fields.len() != 2 {
+                return Err(Error::IllegalArgument {
+                    message: format!(
+                        "Map entries Struct must have 2 fields (key, value), got {}",
+                        fields.len()
+                    ),
+                });
+            }
+            DataTypes::map(from_arrow_field(&fields[0])?, from_arrow_field(&fields[1])?)
+        }
         ArrowDataType::Struct(fields) => {
             let row_fields: Result<Vec<DataField>> = fields
                 .iter()
-                .map(|f| {
-                    let mut dt = from_arrow_type(f.data_type())?;
-                    if !f.is_nullable() {
-                        dt = dt.as_non_nullable();
-                    }
-                    Ok(DataField::new(f.name(), dt, None))
-                })
+                .map(|f| Ok(DataField::new(f.name(), from_arrow_field(f)?, None)))
                 .collect();
             DataTypes::row(row_fields?)
         }
@@ -1848,6 +1871,33 @@ mod tests {
             );
         } else {
             panic!("Expected ArrowDataType::Map, got {:?}", arrow_type);
+        }
+    }
+
+    #[test]
+    fn test_from_arrow_type_preserves_container_field_nullability() {
+        let arrow_list = ArrowDataType::List(Arc::new(arrow_schema::Field::new(
+            "item",
+            ArrowDataType::Int32,
+            false,
+        )));
+        match from_arrow_type(&arrow_list).unwrap() {
+            DataType::Array(at) => assert!(!at.get_element_type().is_nullable()),
+            other => panic!("expected Array, got {other:?}"),
+        }
+
+        let entries_struct = ArrowDataType::Struct(arrow_schema::Fields::from(vec![
+            arrow_schema::Field::new("key", ArrowDataType::Utf8, false),
+            arrow_schema::Field::new("value", ArrowDataType::Int32, false),
+        ]));
+        let entries_field = arrow_schema::Field::new("entries", entries_struct, false);
+        let arrow_map = ArrowDataType::Map(Arc::new(entries_field), false);
+        match from_arrow_type(&arrow_map).unwrap() {
+            DataType::Map(m) => {
+                assert!(!m.key_type().is_nullable());
+                assert!(!m.value_type().is_nullable());
+            }
+            other => panic!("expected Map, got {other:?}"),
         }
     }
 
--- a/crates/fluss/src/row/binary/binary_writer.rs
+++ b/crates/fluss/src/row/binary/binary_writer.rs
@@ -18,10 +18,10 @@
 use crate::error::Error::IllegalArgument;
 use crate::error::Result;
 use crate::metadata::{DataType, RowType};
-use crate::row::Datum;
 use crate::row::Decimal;
 use crate::row::binary::BinaryRowFormat;
 use crate::row::datum::{TimestampLtz, TimestampNtz};
+use crate::row::{Datum, FlussArray, FlussMap};
 
 /// Writer to write a composite data format, like row, array,
 #[allow(dead_code)]
@@ -69,7 +69,9 @@ pub trait BinaryWriter {
 
     fn write_timestamp_ltz(&mut self, value: &TimestampLtz, precision: u32);
 
-    fn write_array(&mut self, value: &[u8]);
+    fn write_array(&mut self, value: &FlussArray);
+
+    fn write_map(&mut self, value: &FlussMap);
 
     // TODO Row serializer
     // fn write_row(&mut self, pos: i32, value: &InternalRow);
@@ -138,6 +140,7 @@ pub enum InnerValueWriter {
     TimestampNtz(u32), // precision
     TimestampLtz(u32), // precision
     Array,
+    Map,
     Row(NestedRowWriter),
 }
 
@@ -207,13 +210,10 @@ impl InnerValueWriter {
                 Ok(InnerValueWriter::TimestampLtz(t.precision()))
             }
             DataType::Array(_) => Ok(InnerValueWriter::Array),
+            DataType::Map(_) => Ok(InnerValueWriter::Map),
             DataType::Row(row_type) => Ok(InnerValueWriter::Row(NestedRowWriter::from_row_type(
                 row_type,
             )?)),
-            _ => unimplemented!(
-                "ValueWriter for DataType {:?} is currently not implemented",
-                data_type
-            ),
         }
     }
     pub fn write_value<W: BinaryWriter>(
@@ -273,7 +273,10 @@ impl InnerValueWriter {
                 writer.write_timestamp_ltz(ts, *p);
             }
             (InnerValueWriter::Array, Datum::Array(arr)) => {
-                writer.write_array(arr.as_bytes());
+                writer.write_array(arr);
+            }
+            (InnerValueWriter::Map, Datum::Map(map)) => {
+                writer.write_map(map);
             }
             (InnerValueWriter::Row(nested_writer), Datum::Row(inner_row)) => {
                 use crate::row::compacted::CompactedRowWriter;
--- a/crates/fluss/src/row/binary/iceberg_binary_row_writer.rs
+++ b/crates/fluss/src/row/binary/iceberg_binary_row_writer.rs
@@ -21,6 +21,8 @@ use crate::error::{Error, Result};
 use crate::metadata::DataType;
 use crate::row::Decimal;
 use crate::row::binary::{BinaryWriter, ValueWriter};
+use crate::row::binary_array::FlussArray;
+use crate::row::binary_map::FlussMap;
 
 const MICROS_PER_MILLI: i64 = 1_000;
 
@@ -225,8 +227,12 @@ impl BinaryWriter for IcebergBinaryRowWriter {
         self.write_raw(&micros.to_le_bytes());
     }
 
-    fn write_array(&mut self, _value: &[u8]) {
-        panic!("Iceberg key columns do not support array values");
+    fn write_array(&mut self, _value: &FlussArray) {
+        unreachable!("Array/Map types are rejected during value writer creation");
+    }
+
+    fn write_map(&mut self, _value: &FlussMap) {
+        unreachable!("Array/Map types are rejected during value writer creation");
     }
 
     fn complete(&mut self) {
--- a/crates/fluss/src/row/binary_array.rs
+++ b/crates/fluss/src/row/binary_array.rs
@@ -30,6 +30,7 @@ use crate::metadata::{DataType, RowType};
 use crate::row::Decimal;
 use crate::row::InternalRow;
 use crate::row::binary::{BinaryRowFormat, ValueWriter};
+use crate::row::binary_map::FlussMap;
 use crate::row::compacted::{CompactedRow, CompactedRowWriter, calculate_bit_set_width_in_bytes};
 use crate::row::datum::{Date, Time, TimestampLtz, TimestampNtz};
 use crate::row::field_getter::FieldGetter;
@@ -74,6 +75,22 @@ pub fn calculate_fix_length_part_size(element_type: &DataType) -> usize {
 /// Matches Java's `roundNumberOfBytesToNearestWord`.
 fn round_to_nearest_word(num_bytes: usize) -> usize {
     (num_bytes + 7) & !7
+}
+
+fn is_variable_length_type(dt: &DataType) -> bool {
+    match dt {
+        DataType::Char(_)
+        | DataType::String(_)
+        | DataType::Binary(_)
+        | DataType::Bytes(_)
+        | DataType::Array(_)
+        | DataType::Map(_)
+        | DataType::Row(_) => true,
+        DataType::Decimal(d) => !Decimal::is_compact_precision(d.precision()),
+        DataType::Timestamp(t) => !TimestampNtz::is_compact(t.precision()),
+        DataType::TimestampLTz(t) => !TimestampLtz::is_compact(t.precision()),
+        _ => false,
+    }
 }
 
 /// A Fluss binary array, wire-compatible with Java's `BinaryArray`.
@@ -217,6 +234,33 @@ impl FlussArray {
         let byte_index = pos >> 3;
         let bit = pos & 7;
         (self.data[4 + byte_index] & (1u8 << bit)) != 0
+    }
+
+    /// Returns the logically occupied bytes of this array, including the variable-length part.
+    /// This is used to detect trailing garbage in binary containers.
+    pub fn extent(&self, element_type: &DataType) -> Result<usize> {
+        let header_size = calculate_header_in_bytes(self.size);
+        let element_size = calculate_fix_length_part_size(element_type);
+        let fixed_part_size = round_to_nearest_word(header_size + self.size * element_size);
+
+        if !is_variable_length_type(element_type) {
+            return Ok(fixed_part_size);
+        }
+
+        let mut max_extent = fixed_part_size;
+        for i in 0..self.size {
+            if !self.is_null_at(i) {
+                let packed = self.read_i64(i, "extent calculation")? as u64;
+                let mark = packed & HIGHEST_FIRST_BIT;
+                if mark == 0 {
+                    let offset = (packed >> 32) as usize;
+                    let len = (packed & 0xFFFF_FFFF) as usize;
+                    max_extent = max_extent.max(offset + len);
+                }
+            }
+        }
+
+        Ok(round_to_nearest_word(max_extent))
     }
 
     fn checked_slice(&self, start: usize, len: usize, context: &str) -> Result<&[u8]> {
@@ -421,6 +465,16 @@ impl FlussArray {
     pub fn get_array(&self, pos: usize) -> Result<FlussArray> {
         let (start, len) = self.read_var_len_span(pos)?;
         FlussArray::from_owned_bytes(self.data.slice(start..start + len))
+    }
+
+    pub fn get_map(
+        &self,
+        pos: usize,
+        key_type: &DataType,
+        value_type: &DataType,
+    ) -> Result<FlussMap> {
+        let (start, len) = self.read_var_len_span(pos)?;
+        FlussMap::from_owned_bytes(self.data.slice(start..start + len), key_type, value_type)
     }
 
     pub fn get_row<'a>(&'a self, pos: usize, row_type: &'a RowType) -> Result<CompactedRow<'a>> {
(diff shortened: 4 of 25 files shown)


leekeiabstraction:
@charlesdong1991 @fresh-borzoni Appreciate if you guys can have a look as well.

crates/fluss/src/row/binary/binary_writer.rs
```diff
@@ -69,6 +69,8 @@ pub trait BinaryWriter {
 
     fn write_array(&mut self, value: &[u8]);
 
+    fn write_map(&mut self, value: &[u8]);
+
```
leekeiabstraction, line 75: Appreciate that you are following convention from `write_array`.

I'm thinking about the ease of read/use of this signature. The caller would have to know in advance how to get the map into the form of a single byte array. 

Why not pass `FlussMap`? Save some `.as_bytes()` calls and get the benefit of stronger compile time check vs. passing byte array.

leekeiabstraction (commented):
TY for the PR! Left an early comment.

crates/fluss/src/row/binary/binary_writer.rs
qzyu999: Hi @leekeiabstraction, thanks for the suggestion, made appropriate changes in 9f995f8ea24873b034284177f530147c60049a12.

crates/fluss/src/row/column.rs
```diff
@@ -433,6 +433,62 @@ impl InternalRow for ColumnarRow {
         write_arrow_values_to_fluss_array(&*values, &element_fluss_type, &mut writer)?;
         writer.complete()
     }
+
+    fn get_map(&self, pos: usize) -> Result<crate::row::binary_map::FlussMap> {
+        use crate::record::from_arrow_type;
+        use crate::row::binary_array::FlussArrayWriter;
+        use crate::row::binary_map::FlussMap;
+        use arrow::array::MapArray;
```
leekeiabstraction: Let's follow current code convention and have `use` at the top of the file.

crates/fluss/src/row/column.rs
leekeiabstraction: Let's follow current code convention and have `use` at the top of the file.

crates/fluss/src/row/column.rs
```diff
@@ -433,6 +433,62 @@ impl InternalRow for ColumnarRow {
         write_arrow_values_to_fluss_array(&*values, &element_fluss_type, &mut writer)?;
         writer.complete()
     }
+
+    fn get_map(&self, pos: usize) -> Result<crate::row::binary_map::FlussMap> {
```
leekeiabstraction: We can make this less verbose by declaring `use` at the top of the file

```rust
fn get_map(&self, pos: usize) -> Result<FlussMap>
```

crates/fluss/src/row/column.rs
leekeiabstraction: We can make this less verbose by declaring `use` at the top of the file

```rust
fn get_array(&self, pos: usize) -> Result<FlussArray>
```

crates/fluss/src/row/binary/binary_writer.rs
```diff
@@ -67,7 +67,9 @@ pub trait BinaryWriter {
 
     fn write_timestamp_ltz(&mut self, value: &crate::row::datum::TimestampLtz, precision: u32);
 
-    fn write_array(&mut self, value: &[u8]);
+    fn write_array(&mut self, value: &crate::row::FlussArray);
+
+    fn write_map(&mut self, value: &crate::row::FlussMap);
```
leekeiabstraction: We can make this less verbose by declaring `use` at the top of the file.

```rust
    fn write_array(&mut self, value: &FlussArray);
    fn write_map(&mut self, value: &FlussMap);
```

crates/fluss/src/row/binary/iceberg_binary_row_writer.rs
```diff
@@ -225,10 +225,14 @@ impl BinaryWriter for IcebergBinaryRowWriter {
         self.write_raw(&micros.to_le_bytes());
     }
 
-    fn write_array(&mut self, _value: &[u8]) {
+    fn write_array(&mut self, _value: &crate::row::FlussArray) {
         panic!("Iceberg key columns do not support array values");
     }
 
+    fn write_map(&mut self, _value: &crate::row::FlussMap) {
```
leekeiabstraction: We can make these signatures less verbose by declaring `use` at the top of the file.

crates/fluss/src/row/compacted/compacted_key_writer.rs
```diff
@@ -111,7 +111,9 @@ impl BinaryWriter for CompactedKeyWriter {
 
             fn write_timestamp_ltz(&mut self, value: &crate::row::datum::TimestampLtz, precision: u32);
 
-            fn write_array(&mut self, value: &[u8]);
+            fn write_array(&mut self, value: &crate::row::FlussArray);
+
+            fn write_map(&mut self, value: &crate::row::FlussMap);
```
leekeiabstraction: We can make these signatures less verbose by declaring `use` at the top of the file.

crates/fluss/src/row/compacted/compacted_row.rs
```diff
@@ -171,6 +171,10 @@ impl<'a> InternalRow for CompactedRow<'a> {
         self.decoded_row()?.get_array(pos)
     }
 
+    fn get_map(&self, pos: usize) -> Result<crate::row::FlussMap> {
```
leekeiabstraction: We can make these signatures less verbose by declaring `use` at the top of the file.

crates/fluss/src/row/compacted/compacted_row.rs
```diff
@@ -491,4 +495,167 @@ mod tests {
         assert_eq!(nested2.size(), 1);
         assert_eq!(nested2.get_int(0).unwrap(), 99);
     }
+
+    #[test]
+    fn test_compacted_row_map() {
+        use crate::metadata::DataTypes;
+        use crate::row::binary_array::FlussArrayWriter;
+        use crate::row::binary_map::FlussMap;
```
leekeiabstraction: These `use` are repeated across test cases. Let's declare them under `mod tests`?

crates/fluss/src/row/compacted/compacted_row.rs
```diff
@@ -491,4 +495,167 @@ mod tests {
...
+        assert_eq!(read_map.size(), 2);
+        assert_eq!(read_map.key_array().get_int(0).unwrap(), 1);
+        assert_eq!(read_map.value_array().get_string(0).unwrap(), "a");
+    }
+
+    #[test]
+    fn test_compacted_row_map_with_nulls() {
+        use crate::metadata::DataTypes;
+        use crate::row::binary_array::FlussArrayWriter;
+        use crate::row::binary_map::FlussMap;
```
leekeiabstraction: These `use` are repeated across test cases. Let's declare them under `mod tests`?

crates/fluss/src/row/compacted/compacted_row.rs
```diff
@@ -491,4 +495,167 @@ mod tests {
...
+        assert_eq!(read_map.size(), 1);
+        assert_eq!(read_map.key_array().get_int(0).unwrap(), 7);
+        assert_eq!(read_map.value_array().get_string(0).unwrap(), "hello");
+    }
+
+    #[test]
+    fn test_compacted_row_nested_map() {
+        use crate::metadata::DataTypes;
+        use crate::row::binary_array::FlussArrayWriter;
+        use crate::row::binary_map::FlussMap;
```
leekeiabstraction: These `use` are repeated across test cases. Let's declare them under `mod tests`?

crates/fluss/src/row/compacted/compacted_row_writer.rs
```diff
@@ -165,8 +165,12 @@ impl BinaryWriter for CompactedRowWriter {
         self.write_bytes(&bytes[..length.min(bytes.len())])
     }
 
-    fn write_array(&mut self, value: &[u8]) {
-        self.write_bytes(value)
+    fn write_array(&mut self, value: &crate::row::FlussArray) {
+        self.write_bytes(value.as_bytes())
+    }
+
+    fn write_map(&mut self, value: &crate::row::FlussMap) {
```
leekeiabstraction: Ditto on more succinct signature by declaring `use` at top.

crates/fluss/src/row/binary_array.rs
```diff
@@ -688,6 +699,10 @@ impl crate::row::InternalRow for FlussArray {
     fn get_array(&self, pos: usize) -> Result<FlussArray> {
         self.get_array(pos)
     }
+
+    fn get_map(&self, pos: usize) -> Result<crate::row::FlussMap> {
```
leekeiabstraction: Ditto on more succinct signature by declaring `use` at top.

crates/fluss/src/row/column_writer.rs
```diff
@@ -720,6 +823,45 @@ fn finish_list_array(
...
+    keys: ArrayRef,
+    values: ArrayRef,
+    value_nullable: bool,
+    offsets: &[i32],
+    validity: &[bool],
+) -> ArrayRef {
+    use arrow::array::{Array, MapArray, StructArray};
+    use arrow::buffer::{NullBuffer, OffsetBuffer, ScalarBuffer};
+    use arrow::datatypes::Field;
+    use std::sync::Arc;
```
leekeiabstraction, line 966: These seems fine as it's only really used within `finish_map_array`.

crates/fluss/src/row/column_writer.rs
```diff
@@ -993,4 +1135,54 @@ mod tests {
             "Arrow field inside the list should be non-nullable"
         );
     }
+
+    #[test]
+    fn test_write_map_type() {
+        use crate::metadata::DataTypes;
+        use crate::row::binary_array::FlussArrayWriter;
+        use crate::row::binary_map::FlussMap;
```
leekeiabstraction: These seems fine as it's only really used within `test_write_map_type`.

crates/fluss/src/row/binary_map.rs
```diff
@@ -0,0 +1,331 @@
...
+        let mut value_writer = FlussArrayWriter::new(2, &DataTypes::string());
+        value_writer.write_string(0, "a");
+        value_writer.write_string(1, "b");
+        let value_array = value_writer.complete().unwrap();
+
+        let map = FlussMap::from_arrays(&key_array, &value_array).unwrap();
+        assert_eq!(map.size(), 2);
+
+        let bytes = map.as_bytes();
+        let decoded = FlussMap::from_bytes(bytes).unwrap();
```
leekeiabstraction: This tests the Rust core side round trip serde correctness. Let's add an assertion on absolute bytes value, preferably copied from Java test case side to ensure serde correctness across Rust core and Java side.

crates/fluss/src/row/binary_map.rs
```diff
@@ -0,0 +1,331 @@
...
+        let key_array = key_writer.complete().unwrap();
+
+        let mut value_writer = FlussArrayWriter::new(3, &DataTypes::int());
+        value_writer.write_int(0, 10);
+        value_writer.set_null_at(1); // null value
+        value_writer.write_int(2, 30);
+        let value_array = value_writer.complete().unwrap();
+
+        let map = FlussMap::from_arrays(&key_array, &value_array).unwrap();
+        let decoded = FlussMap::from_bytes(map.as_bytes()).unwrap();
```
leekeiabstraction: This tests the Rust core side round trip serde correctness. Let's add an assertion on absolute bytes value, preferably copied from Java test case side to ensure serde correctness across Rust core and Java side.

crates/fluss/src/row/binary_map.rs
```diff
@@ -0,0 +1,331 @@
...
+
+    #[test]
+    fn test_mismatched_array_sizes() {
+        let key_writer = FlussArrayWriter::new(1, &DataTypes::int());
+        let key_array = key_writer.complete().unwrap();
+
+        let value_writer = FlussArrayWriter::new(2, &DataTypes::string());
+        let value_array = value_writer.complete().unwrap();
+
+        assert!(FlussMap::from_arrays(&key_array, &value_array).is_err());
```
leekeiabstraction: What's the error message in this case?

crates/fluss/src/row/binary_map.rs
```diff
@@ -0,0 +1,331 @@
...
+        bad_data.extend_from_slice(&neg_size);
+        bad_data.extend_from_slice(&[0, 0, 0, 0]);
+        assert!(FlussMap::from_bytes(&bad_data).is_err());
+
+        // Key array length exceeds payload
+        let large_size = 100i32.to_le_bytes();
+        let mut bad_data2 = vec![];
+        bad_data2.extend_from_slice(&large_size);
+        bad_data2.extend_from_slice(&[0, 0, 0, 0]);
+        assert!(FlussMap::from_bytes(&bad_data2).is_err());
```
leekeiabstraction: What's the error message in this case?

crates/fluss/src/row/column.rs
```diff
@@ -433,6 +433,62 @@ impl InternalRow for ColumnarRow {
...
+        let mut key_writer = FlussArrayWriter::new(len, &key_fluss_type);
+        write_arrow_values_to_fluss_array(&**keys_arrow, &key_fluss_type, &mut key_writer)?;
+        let key_array = key_writer.complete()?;
+
+        // Convert Arrow values → FlussArray
+        let mut value_writer = FlussArrayWriter::new(len, &value_fluss_type);
+        write_arrow_values_to_fluss_array(&**values_arrow, &value_fluss_type, &mut value_writer)?;
+        let value_array = value_writer.complete()?;
+
+        FlussMap::from_arrays(&key_array, &value_array)
```
leekeiabstraction: These lines which parse from `StructArray` to `FlussMap` seem duplicated, let's apply DRY and refactor.

crates/fluss/src/row/column.rs
```diff
@@ -653,6 +709,53 @@ fn write_arrow_values_to_fluss_array(
...
+                    let mut val_w = FlussArrayWriter::new(entry_len, &value_fluss_type);
+                    write_arrow_values_to_fluss_array(
+                        &**values_arrow,
+                        &value_fluss_type,
+                        &mut val_w,
+                    )?;
+                    let value_array = val_w.complete()?;
+
+                    let fluss_map =
+                        crate::row::binary_map::FlussMap::from_arrays(&key_array, &value_array)?;
```
leekeiabstraction: These lines which parse from `StructArray` to `FlussMap` seem duplicated, let's apply DRY and refactor.

crates/fluss/src/row/binary_map.rs
```diff
@@ -0,0 +1,331 @@
...
+        let (key_array, value_array) = Self::validate(data)?;
+        Ok(FlussMap {
+            data: Bytes::copy_from_slice(data),
+            key_array,
+            value_array,
+        })
+    }
+
+    /// Creates a FlussMap from owned bytes without copying.
+    pub fn from_owned_bytes(data: Bytes) -> Result<Self> {
```
leekeiabstraction: Do these need to be `pub`? Can they be `pub(crate)`?

crates/fluss/src/row/binary_map.rs
```diff
@@ -0,0 +1,331 @@
...
+            data,
+            key_array,
+            value_array,
+        })
+    }
+
+    /// Creates a FlussMap by combining a key array and a value array.
+    ///
+    /// Copies both arrays into a new contiguous buffer.
+    pub fn from_arrays(key_array: &FlussArray, value_array: &FlussArray) -> Result<Self> {
```
leekeiabstraction, line 213: IIUC, this is the main method through which we expect end user to use `FlussMap`. This method signature is rather clunky for end user to use because they need to initialise two `FlussArray`. 

Let's use a more user friendly signature e.g. something that accepts `HashMap`

leekeiabstraction (commented):
TY for the PR! Added comments, PTAL @qzyu999

fresh-borzoni (commented):
@qzyu999 Ty for the PR, GH is acting weird, so I'll paste my review like this:

crates/fluss/src/row/column_writer.rs:563(append_null)
When Datum::Null is appended to a Map column, the path falls through with_builder! and crashes at runtime. 
Probably we need an arm for Map as well

crates/fluss/src/row/binary_map.rs:172 and probably in datatype.rs
Shall we check that key is not null?

crates/fluss/src/row/column_writer.rs
```diff
@@ -993,4 +1135,54 @@ mod tests {
             "Arrow field inside the list should be non-nullable"
         );
     }
+
+    #[test]
+    fn test_write_map_type() {
```
charlesdong1991, line 1270: can you add a test where a `NULL` Map value is written?

crates/fluss/src/row/column.rs
charlesdong1991: +1, parsing logic seems very similar if not identical, can be wrapped to a helper function

crates/fluss/src/row/datum.rs
```diff
@@ -550,6 +571,62 @@ fn append_fluss_array_to_list_builder(
     Ok(())
 }
 
+fn append_fluss_map_to_map_builder(
+    map: &crate::row::FlussMap,
+    builder: &mut dyn ArrayBuilder,
+    data_type: &arrow_schema::DataType,
+) -> Result<()> {
+    use crate::record::from_arrow_type;
```
charlesdong1991: Does Arrow type support Map? can we add a test to see if it does or not?

crates/fluss/src/row/binary/iceberg_binary_row_writer.rs
```diff
@@ -225,10 +225,14 @@ impl BinaryWriter for IcebergBinaryRowWriter {
...
+        panic!("Iceberg key columns do not support map values");
```
charlesdong1991: With a second thought now, i feel it might be better to return an error, instead of panicing, the same applies to array.

crates/fluss/src/row/binary_map.rs
```diff
@@ -0,0 +1,331 @@
...
+    fn test_mismatched_array_sizes() {
+        let key_writer = FlussArrayWriter::new(1, &DataTypes::int());
+        let key_array = key_writer.complete().unwrap();
+
+        let value_writer = FlussArrayWriter::new(2, &DataTypes::string());
+        let value_array = value_writer.complete().unwrap();
+
+        assert!(FlussMap::from_arrays(&key_array, &value_array).is_err());
+    }
+}
```
charlesdong1991, line 561: i wonder if we can add a test for nested Map?

charlesdong1991 (commented):
Thanks for the nice PR 👍  left some comments!

crates/fluss/src/row/column.rs
qzyu999: Hi @leekeiabstraction, this has been addressed in 558b7eeeeb36249e06d37de428cfc4482de7c796.

crates/fluss/src/row/column.rs
qzyu999: Hi @leekeiabstraction, this has been addressed in 558b7eeeeb36249e06d37de428cfc4482de7c796.

crates/fluss/src/row/column.rs
qzyu999: Hi @leekeiabstraction, this has been addressed in 558b7eeeeb36249e06d37de428cfc4482de7c796.

crates/fluss/src/row/column.rs
qzyu999: Hi @leekeiabstraction, this has been addressed in 558b7eeeeb36249e06d37de428cfc4482de7c796.

crates/fluss/src/row/binary/binary_writer.rs
qzyu999: Hi @leekeiabstraction, this has been addressed in 558b7eeeeb36249e06d37de428cfc4482de7c796.

crates/fluss/src/row/binary/iceberg_binary_row_writer.rs
qzyu999: Hi @leekeiabstraction, this has been addressed in 94fcceaa1ff3659772705f40162e53b20349941f.

crates/fluss/src/row/compacted/compacted_key_writer.rs
qzyu999: Hi @leekeiabstraction, this has been addressed in 94fcceaa1ff3659772705f40162e53b20349941f.

crates/fluss/src/row/compacted/compacted_row.rs
qzyu999: Hi @leekeiabstraction, this has been addressed in 94fcceaa1ff3659772705f40162e53b20349941f.

crates/fluss/src/row/compacted/compacted_row.rs
qzyu999: Hi @leekeiabstraction, this has been addressed in 94fcceaa1ff3659772705f40162e53b20349941f.

crates/fluss/src/row/compacted/compacted_row.rs
qzyu999: Hi @leekeiabstraction, this has been addressed in 94fcceaa1ff3659772705f40162e53b20349941f.

crates/fluss/src/row/compacted/compacted_row.rs
qzyu999: Hi @leekeiabstraction, this has been addressed in 94fcceaa1ff3659772705f40162e53b20349941f.

crates/fluss/src/row/compacted/compacted_row_writer.rs
qzyu999: Hi @leekeiabstraction, this has been addressed in 94fcceaa1ff3659772705f40162e53b20349941f.

crates/fluss/src/row/binary_array.rs
qzyu999: Hi @leekeiabstraction, this has been addressed in 94fcceaa1ff3659772705f40162e53b20349941f.

crates/fluss/src/row/binary_map.rs
qzyu999: Hi @charlesdong1991, this has been addressed in 471016f095aafac0cac471d9e1fd06429ab677a0.

qzyu999:
> crates/fluss/src/row/binary_map.rs:172 ... Shall we check that key is not null?

Hi @fresh-borzoni, this has been addressed in 471016f095aafac0cac471d9e1fd06429ab677a0.

crates/fluss/src/row/binary_map.rs
qzyu999: Hi @leekeiabstraction, this has been addressed in 471016f095aafac0cac471d9e1fd06429ab677a0.

crates/fluss/src/row/binary_map.rs
qzyu999: Hi @leekeiabstraction, this has been addressed in 471016f095aafac0cac471d9e1fd06429ab677a0.

crates/fluss/src/row/binary_map.rs
qzyu999: Hi @leekeiabstraction, this has been addressed in 471016f095aafac0cac471d9e1fd06429ab677a0.

crates/fluss/src/row/binary_map.rs
qzyu999: Hi @leekeiabstraction, this has been addressed in 471016f095aafac0cac471d9e1fd06429ab677a0.

crates/fluss/src/row/binary_map.rs
qzyu999: Hi @leekeiabstraction, this has been addressed in 471016f095aafac0cac471d9e1fd06429ab677a0.

crates/fluss/src/row/binary_map.rs
qzyu999: Hi @leekeiabstraction, thanks for your reply I think you bring up a valid point.

My understanding is that because `FlussMap` writes to a strict binary layout, we need the exact `DataType` schema (like `DataTypes::int()`) at runtime to initialize the `FlussArrayWriter`s.

If we tried to accept a standard `HashMap<K, V>`, we wouldn't have that Fluss schema metadata inside the method to know how to properly pack the bytes. Because of that, it seemed like `from_arrays` had to be the required low-level API since the caller has already explicitly built the arrays with the correct schemas.

IIUC, it would make sense to keep `from_arrays` for the core API, but maybe add a helper method in `test_utils` like `from_hashmap(map, key_schema, val_schema)` to make the test UX nicer? Please let me know what your thoughts are about this.

crates/fluss/src/row/column_writer.rs
qzyu999: Hi @charlesdong1991, this has been addressed in 21e8ca974bc8ac3f83f887199d2e18c795fa8fbd.

qzyu999:
> crates/fluss/src/row/column_writer.rs:563(append_null) When Datum::Null is appended to a Map column, the path falls through with_builder! and crashes at runtime. Probably we need an arm for Map as well

Hi @fresh-borzoni, this has been addressed in 21e8ca974bc8ac3f83f887199d2e18c795fa8fbd.

crates/fluss/src/row/datum.rs
qzyu999: Hi @charlesdong1991, this has been addressed in fa3adfa0ec5510187bd1c204f605e1c359b5852f.

crates/fluss/src/row/binary/iceberg_binary_row_writer.rs
qzyu999: Hi @charlesdong1991, I addressed this in 94fcceaa1ff3659772705f40162e53b20349941f where I return an `unreachable!()` instead, which I believe should work since the function returns void. To return an error my understanding is that it would require rewriting the `BinaryWriter` trait which would lead to a much larger refactor of all other code that touches it.

crates/fluss/src/row/column.rs
qzyu999: Hi @leekeiabstraction and @charlesdong1991, this has been addressed in 558b7eeeeb36249e06d37de428cfc4482de7c796.

crates/fluss/src/row/column.rs
qzyu999: Hi @leekeiabstraction, this has been addressed in 558b7eeeeb36249e06d37de428cfc4482de7c796.

qzyu999:
> ... probably in datatype.rs

Hi @fresh-borzoni, this has been addressed in 4c06cd971a4d9b3a488ca71846fcc48c001e28a4.

qzyu999:
Hi @leekeiabstraction, @fresh-borzoni, and @charlesdong1991, thank you all for the review suggestions. I've replied to all of your comments and have made corresponding changes, PTAL.

crates/fluss/src/row/binary/iceberg_binary_row_writer.rs
```diff
@@ -225,8 +227,12 @@ impl BinaryWriter for IcebergBinaryRowWriter {
         self.write_raw(&micros.to_le_bytes());
     }
 
-    fn write_array(&mut self, _value: &[u8]) {
-        panic!("Iceberg key columns do not support array values");
+    fn write_array(&mut self, _value: &FlussArray) {
+        unreachable!("Array/Map types are rejected during value writer creation");
```
fresh-borzoni, line 231: It's the same panic under the hood

crates/fluss/src/row/binary_map.rs
```diff
@@ -0,0 +1,400 @@
...
+        S: serde::Serializer,
+    {
+        serializer.serialize_bytes(&self.data)
+    }
+}
+
+impl FlussMap {
+    /// Validates the raw bytes and extracts the sub-arrays.
+    fn validate(data: &[u8]) -> Result<(FlussArray, FlussArray)> {
+        if data.len() < 4 {
```
fresh-borzoni, line 114: validate accepts trailing bytes -  FlussArray::from_bytes only checks the header, not that the slice is fully consumed

 Could we assert 4 + key_extent + value_extent == data.len() at the end of validate?

crates/fluss/src/row/binary_map.rs
```diff
@@ -0,0 +1,400 @@
...
+            return Err(IllegalArgument {
+                message: format!(
+                    "FlussMap key array size ({}) does not match value array size ({})",
+                    key_array.size(),
+                    value_array.size()
+                ),
+            });
+        }
+
+        for i in 0..key_array.size() {
```
fresh-borzoni: The null-key check is in from_arrays but not from_bytes. With non-nullable keys, column_writerskips the null check and writes garbage to Arrow.

Could the check move into validate so all paths share it?

crates/fluss/src/record/arrow.rs
fresh-borzoni, line 1238: Arrow spec requires the entries field of Map to be non-nullable
Should be Field::new("entries", ..., false).

crates/fluss/src/row/datum.rs
```diff
@@ -550,6 +571,62 @@ fn append_fluss_array_to_list_builder(
...
+            }
+        },
+        _ => {
+            return Err(RowConvertError {
+                message: format!("Expected Map Arrow type for Map datum, got: {data_type:?}"),
+            });
+        }
+    };
+
+    let key_fluss_type = from_arrow_type(&key_arrow_type)?;
```
fresh-borzoni: Key/value Fluss types come from the Arrow type, not the FlussMap. What if is they mismatch?

fresh-borzoni (commented):
@qzyu999 Ty for the changes, looked through, looks good, left comments, PTAL

Also can we add simple e2e integration test to see if everything is wired correctly?
Scenario:
  - Create a table with one Map column                                                                                                                                                                                                               
  - Insert one row                                                                                                                                                                                                                                   
  - Fetch it back 
  - Assert the map round-trips (size + at least one key/value)

crates/fluss/src/row/binary_map.rs
leekeiabstraction: Hello @qzyu999 , 

I realise I should elaborate more. Suppose a client wants to set a map field on generic row with current pattern, if I understand correctly the pattern that they can use with current shape of API is below:

```rust
let mut row = GenericRow::new(2);
row.set_field(0, i as i32);
let keyWriter = FlussArrayWriter::new(2, &DataType::String());
keyWriter.write_string(0,"hello");
keyWriter.write_string(1,"bonjour");
let valueWriter = FlussArrayWriter::new(2, &DataType::Int());
valueWriter.write_int(0,1);
valueWriter.write_int(1,2);
row.set_field(1, FlussMap::from_arrays(keyWriter.complete()?, valueWriter.complete()?));
```

This is rather verbose and error prone. We can instead we have a `FlussMapWriter`, that allows clients to use a more Map like write pattern e.g. `mapWriter.write_entry(A, B)` or `mapWriter.write_map(map)`. Type checking, conversion to FlussArray etc. can happen within FlussMapWriter.

crates/fluss/src/row/binary/iceberg_binary_row_writer.rs
qzyu999: Hi @fresh-borzoni, I responded to @charlesdong1991 regarding this with the following:

> Hi @charlesdong1991, I addressed this in https://github.com/apache/fluss-rust/commit/94fcceaa1ff3659772705f40162e53b20349941f where I return an unreachable!() instead, which I believe should work since the function returns void. To return an error my understanding is that it would require rewriting the BinaryWriter trait which would lead to a much larger refactor of all other code that touches it.

fresh-borzoni: Sorry, I missed this, let's leave like this then

crates/fluss/src/record/arrow.rs
qzyu999: Hi @fresh-borzoni, this has been addressed in c3d032acf759248b91c5a5c779f7ebb8c20777f8.

crates/fluss/src/row/datum.rs
qzyu999: Hi @fresh-borzoni, this has been addressed in c3d032acf759248b91c5a5c779f7ebb8c20777f8.

crates/fluss/src/row/binary_map.rs
qzyu999: Hi @fresh-borzoni, addressed in 25b2f571fbb211786a4fb6bca832702db3bc00d5.

crates/fluss/src/row/binary_map.rs
qzyu999: Hi @fresh-borzoni, addressed in 25b2f571fbb211786a4fb6bca832702db3bc00d5.

qzyu999:
> @qzyu999 Ty for the changes, looked through, looks good, left comments, PTAL
> 
> Also can we add simple e2e integration test to see if everything is wired correctly? Scenario:
> 
> * Create a table with one Map column
> * Insert one row
> * Fetch it back
> * Assert the map round-trips (size + at least one key/value)

Hi @fresh-borzoni, thanks for the review. I've pushed a series of commits addressing all your feedback:

- Arrow Spec & Types: Hardcoded the `entries` field to be non-nullable and updated the extraction logic to validate against the authoritative client schema.

- Binary Validation: Updated `FlussMap::validate` to enforce strict bounds checking (accounting for 8-byte padding) and centralized the null-key check.

- E2E Testing: Added `test_map_datatype_roundtrip` in `log_table.rs` which verifies the full serialization/deserialization cycle across the network.

PTAL when you have a chance.

crates/fluss/src/row/binary_map.rs
qzyu999: Hi @leekeiabstraction, thanks for clarifying. d962fb6e184068a85ec5a11f1b6a5b2bdb40ac90 introduces `FlussMapWriter`. It completely encapsulates the parallel `FlussArrayWriters`, handles the index tracking, enforces the null-key constraints, and exposes a clean `write_entry(key: Datum, value: Datum)` API.

I also went through and refactored the test suites (including `binary_map`, `datum`, and the integration tests) to use it, and it significantly cuts down on the boilerplate.

fresh-borzoni (approved):
@qzyu999 Ty for the change, LGTM overall 👍 

Would you mind to file a couple of followups: docs and integration tests suite for this? 

@charlesdong1991 @leekeiabstraction Can you take a look?

charlesdong1991 (approved):
thanks for great PR! @qzyu999 overall looks good, and we can use follow-up PRs to address minor issues if seen!

fresh-borzoni:
I think we also need to do the same for Array/Map dts: https://github.com/apache/fluss-rust/issues/543, mb it's a better idea to actually handle them together in single refactor, then fix FFI layer in bindings and finish with Map/Row bindings after this.

qzyu999:
Hi @fresh-borzoni, just merged the branch with latest from main, there were some big changes from d4a0bb6c784568183f6ddac10a7ba493fb37cfb4. Tests are passing, but it would be good for further review. CC: @hemanthsavasere

fresh-borzoni:
@qzyu999 Ty for the great work, I pushed some minor fixes, merging ...

PR #531, opened 2026-05-06
[rust] Support API_VERSION in rust client.

### LINKED ISSUE #516: Support API_VERSIONS in rust client
### Search before asking

### Motivation

Currently, fluss rust client send any request as ApiVersion(0). However, some api versions in latest server has already upgraded to version 1. For example:
```java
// Version 0: Uses lake's encoder for primary key encoding (legacy behavior).
    // Version 1: Uses CompactedKeyEncoder for primary key encoding when bucket key differs from
    //            primary key, enabling prefix lookup support.
    PUT_KV(1016, 0, 1, PUBLIC),

    // Version 0: Uses lake's encoder for primary key encoding (legacy behavior).
    // Version 1: Uses CompactedKeyEncoder for primary key encoding when bucket key differs from
    //            primary key, enabling prefix lookup support.
    LOOKUP(1017, 0, 1, PUBLIC),

// Version 0: Uses lake's encoder for prefix key encoding (legacy behavior).
    // Version 1: Uses CompactedKeyEncoder for prefix key encoding when bucket key differs from
    //            primary key, ensuring encoded bucket key bytes are a prefix of primary key bytes.
    PREFIX_LOOKUP(1034, 0, 1, PUBLIC),

```

Thus, if use rust client read or write to lake table, exception will occurs.

### Solution

_No response_

### Anything else?

_No response_

### Willingness to contribute

loserwang1024:
### Purpose

Linked issue: close #516

### DIFF
--- a/crates/fluss/src/proto/fluss_api.proto
+++ b/crates/fluss/src/proto/fluss_api.proto
@@ -24,6 +24,23 @@ message ErrorResponse {
   optional string error_message = 2;
 }
 
+// api versions request and response
+message ApiVersionsRequest {
+  required string client_software_name = 1;
+  required string client_software_version = 2;
+}
+
+message ApiVersionsResponse {
+  repeated PbApiVersion api_versions = 1;
+  optional int32 server_type = 2;
+}
+
+message PbApiVersion {
+  required int32 api_key = 1;
+  required int32 min_version = 2;
+  required int32 max_version = 3;
+}
+
 // metadata request and response, request send from client to each server.
 message MetadataRequest {
   repeated PbTablePath table_path = 1;
--- a/crates/fluss/src/rpc/server_connection.rs
+++ b/crates/fluss/src/rpc/server_connection.rs
@@ -15,7 +15,7 @@
 // specific language governing permissions and limitations
 // under the License.
 
-use crate::cluster::ServerNode;
+use crate::cluster::{ServerNode, ServerType};
 use crate::error::Error;
 use crate::metrics::{
     CLIENT_BYTES_RECEIVED_TOTAL, CLIENT_BYTES_SENT_TOTAL, CLIENT_REQUEST_LATENCY_MS,
@@ -88,14 +88,14 @@ impl ServerApiVersions {
     pub fn new(server_versions: &[PbApiVersion]) -> Self {
         let mut versions = HashMap::new();
         for sv in server_versions {
-            let api_key = ApiKey::from(sv.api_key as i16);
+            let api_key = ApiKey::from(i16::try_from(sv.api_key).unwrap());
             // Skip unknown API keys — the client does not support them.
             let client_range = match api_key.supported_versions() {
                 Some(range) => range,
                 None => continue,
             };
-            let server_min = sv.min_version as i16;
-            let server_max = sv.max_version as i16;
+            let server_min = i16::try_from(sv.min_version).unwrap();
+            let server_max = i16::try_from(sv.max_version).unwrap();
             let min_version = client_range.min().0.max(server_min);
             let max_version = client_range.max().0.min(server_max);
             if min_version > max_version {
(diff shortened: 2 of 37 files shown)


loserwang1024:
@luoyuxia @fresh-borzoni , CC

crates/fluss/src/proto/fluss_api.proto
```diff
@@ -24,6 +24,23 @@ message ErrorResponse {
...
 
+// api versions request and response
+message ApiVersionsRequest {
+  required string client_software_name = 1;
+  required string client_software_version = 2;
+}
+
+message ApiVersionsResponse {
+  repeated PbApiVersion api_versions = 1;
+  optional int32 server_type = 2;
```
fresh-borzoni, line 35: `server_type` is added but never read. Java validates it and disconnects
  on coordinator/tablet mismatch. Let's file a follow-up, just so we don't forget about it

crates/fluss/src/rpc/server_connection.rs
```diff
@@ -72,6 +74,89 @@ impl fmt::Debug for SaslConfig {
...
+/// The `ApiVersions` request itself always uses `ApiVersion(0)` since it is
+/// sent before the version negotiation handshake.
+/// For all other requests, the handshake must have been completed; otherwise
+/// an error is returned to prevent silent fallback to an incorrect version.
+fn resolve_api_version_for(
+    api_versions: Option<&ServerApiVersions>,
+    api_key: ApiKey,
+) -> Result<ApiVersion, Error> {
+    if api_key == ApiKey::ApiVersion {
+        return Ok(ApiVersion(0));
```
fresh-borzoni: I think Java returns apiKey.highestSupportedVersion, it will desync eventually

crates/fluss/src/rpc/server_connection.rs
```diff
@@ -72,6 +74,89 @@ impl fmt::Debug for SaslConfig {
...
+pub struct ServerApiVersions {
+    versions: HashMap<ApiKey, Result<ApiVersion, String>>,
+}
+
+impl ServerApiVersions {
+    /// Build from the server's advertised API version list.
+    pub fn new(server_versions: &[PbApiVersion]) -> Self {
+        let mut versions = HashMap::new();
+        for sv in server_versions {
+            let api_key = ApiKey::from(i16::try_from(sv.api_key).unwrap());
```
fresh-borzoni, line 91: it can potentially panic, Java truncates to short, shall we do smth similar?

fresh-borzoni (approved):
@loserwang1024 

Ty for the PR, LGTM overall, left a couple of comments, non-blocking
You may wish to push them to followup, if you wish

crates/fluss/src/proto/fluss_api.proto
loserwang1024: add the check and test

crates/fluss/src/rpc/server_connection.rs
loserwang1024: done it.

crates/fluss/src/rpc/server_connection.rs
fresh-borzoni: somehow it got fixed in f7761c4 and then reappeared :)

crates/fluss/src/rpc/server_connection.rs
```diff
@@ -1030,4 +1171,126 @@ mod tests {
...
+
+        assert!(matches!(
+            validate_server_type(
+                &ServerType::CoordinatorServer,
+                Some(ServerType::TabletServer.to_type_id()),
+            ),
+            Err(Error::InvalidServerType { .. })
+        ));
+
+        // Unknown / unmapped type id still fails, with the raw id surfaced so
```
fresh-borzoni, line 1293: None is the skip case, not the unknown case. Probably we want two  assertions: one for None (returns Ok) and one for Some(99) (returns Err)

loserwang1024: done

crates/fluss/src/rpc/server_connection.rs
fresh-borzoni: i filed an issue: https://github.com/apache/fluss-rust/issues/535

fresh-borzoni:
@loserwang1024 Ty, merging

PR #532, opened 2026-05-07
[rust][writer] Add DynamicWriteBatchSizeEstimator

### LINKED ISSUE #445: Implement DynamicWriteBatchSizeEstimator
### Search before asking

### Description

Java dynamically adjusts the batch size target per table based on observed actual batch sizes - grows 10% when batches fill past 80%, shrinks 5% when below 50%, clamped to [2 * pageSize, maxBatchSize]. This right-sizes memory reservations for tables with small rows. We currently use a fixed writer_batch_size for all tables. The impact is less critical for us than Java since our memory limiter is semaphore-based (no physical page pool), but it would still improve memory budget utilization.

### Willingness to contribute

fresh-borzoni:
## Summary

closes #445

Adds DynamicWriteBatchSizeEstimator, mirroring Java Fluss's per-table batch size tuning. When writer.dynamic-batch-size-enabled is on, each table's target grows 10% after a batch fills past 80% and shrinks 5% below 50%, clamped to [writer.batch-size-min, writer.batch-size].
                                                                                                                                                                                                                                                     
Wired into the writer's allocation path: append reads the per-table target instead of the static writer_batch_size; drain feeds the actual size back. The estimator lives on BucketAndWriteBatches behind a parking_lot::Mutex - both touch points short-circuit on the config flag, so the disabled path locks nothing.


fresh-borzoni:
@leekeiabstraction @charlesdong1991 PTAL 🙏

crates/fluss/src/client/write/accumulator.rs
```diff
@@ -938,6 +950,29 @@ struct BucketAndWriteBatches {
...
+}
+
+impl BucketAndWriteBatches {
+    fn new(
+        table_id: TableId,
+        is_partitioned_table: bool,
+        partition_id: Option<PartitionId>,
+        config: &Config,
+    ) -> Self {
+        let estimator = DynamicWriteBatchSizeEstimator::new(
```
charlesdong1991: so this estimator is always being constructed no matter the dynamic setting is true or false?

crates/fluss/src/client/write/accumulator.rs
```diff
@@ -307,28 +308,32 @@ impl RecordAccumulator {
...
             let bucket_and_batches = binding.value_mut();
             let dq = bucket_and_batches
                 .batches
                 .entry(bucket_id)
                 .or_insert_with(|| Arc::new(Mutex::new(VecDeque::new())))
                 .clone();
+            let dynamic_target = self
+                .config
+                .writer_dynamic_batch_size_enabled
+                .then(|| bucket_and_batches.dynamic_batch_size.lock().current());
```
charlesdong1991: given how we use `dynamic_batch_size` here, i wonder if AtomicUsize will be a better option here than Mutex to avoid locking?

website/docs/user-guide/cpp/api-reference.md
```diff
@@ -21,7 +21,9 @@ Complete API reference for the Fluss C++ client.
 | `writer_request_max_size`             | `int32_t`     | `10485760` (10 MB)   | Maximum request size in bytes                                                            |
 | `writer_acks`                         | `std::string` | `"all"`              | Acknowledgment setting (`"all"`, `"0"`, `"1"`, or `"-1"`)                                |
 | `writer_retries`                      | `int32_t`     | `INT32_MAX`          | Number of retries on failure                                                             |
-| `writer_batch_size`                   | `int32_t`     | `2097152` (2 MB)     | Batch size for writes in bytes                                                           |
+| `writer_batch_size`                   | `int32_t`     | `2097152` (2 MB)     | Batch size for writes in bytes (also the upper bound when dynamic sizing is enabled)     |
```
charlesdong1991: Feel a little bit counterintuitive from user perspective to have `writer_batch_size` for max, while `writer_batch_size_min` for min 😅 but very personal feeling.

Nit: on doc itself, maybe clearer to mention it is default batch size when dynamic sizing is disabled?

crates/fluss/src/client/write/accumulator.rs
```diff
@@ -690,6 +692,15 @@ impl RecordAccumulator {
         self.incomplete_batches.write().remove(&batch_id);
     }
 
+    fn record_actual_batch_size(&self, table_path: &Arc<PhysicalTablePath>, actual: usize) {
+        if !self.config.writer_dynamic_batch_size_enabled {
+            return;
+        }
+        if let Some(entry) = self.write_batches.get(table_path) {
+            entry.dynamic_batch_size.lock().update(actual);
```
charlesdong1991: Nit: i wonder in production environment, would it be helpful to see tracing/debugging message when value changes since it might be quite important change that impacts performance?

crates/fluss/src/client/write/accumulator.rs
```diff
@@ -1563,4 +1598,77 @@ mod tests {
             .await
             .expect("notified should complete after wakeup_sender");
     }
+
+    #[test]
+    fn dynamic_batch_size_shrinks_after_small_drained_batch() {
```
charlesdong1991, line 1628: Nit: can we add a complementary test to verify grow direction end-to-end? aka, it shrinks estimate first, so drain small batches, then append that drained batch goes beyond 80% of target, and then we should verify that next allocation will increase?

this is very minor, we can ignore too

charlesdong1991 (commented):
Nice PR!! i left some thoughts! @fresh-borzoni PTAL 🙏

crates/fluss/src/client/write/accumulator.rs
fresh-borzoni: Good point

website/docs/user-guide/cpp/api-reference.md
fresh-borzoni: I would leave `writer_batch_size` as it is as it's consistent with java, just renamed new key a bit

crates/fluss/src/client/write/accumulator.rs
fresh-borzoni: Good idea, ty

fresh-borzoni:
@charlesdong1991 Ty for the review, addressed, PTAL 🙏

charlesdong1991 (approved):
Looks great! 👍 

Can we create a follow-up issue to consider adding an integration test for this optimization?

fresh-borzoni:
Added an issue for integration tests: https://github.com/apache/fluss-rust/issues/539

PR #542, opened 2026-05-10
[test][chore] Capture cluster container logs as CI artifact

charlesdong1991:
### Purpose

Small follow-up from https://github.com/apache/fluss-rust/pull/529, in current timeout failure, if there is fluss cluster container error, we do not see the logs, and this hinders us to understand where it goes wrong for debugging.

So one idea is to save those logs. Here try to capture container logs as CI artifacts that can be downloaded.

This won't affect local development because cluster will be stopped as how it works, but in CI, cluster won't be stopped until pytest finishes, and it should be safe because CI runners are ephemeral VMs, so they should be destroyed after the job.

### DIFF
--- a/.github/workflows/build_and_test_python.yml
+++ b/.github/workflows/build_and_test_python.yml
@@ -81,3 +81,20 @@ jobs:
         env:
           RUST_LOG: DEBUG
           RUST_BACKTRACE: full
+          FLUSS_SKIP_CLUSTER_TEARDOWN: "1"
+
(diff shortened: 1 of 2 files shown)


charlesdong1991:
@fresh-borzoni @leekeiabstraction not sure if this is something we want, but propose here to get some feedback 🙏

fresh-borzoni (approved):
@charlesdong1991 seems to work, let's add this to have more context 👍

PR #544, opened 2026-05-10
[rust] Add end-to-end ROW type

### LINKED ISSUE #388: Row data type support in Rust

fresh-borzoni:
Continues work from #442 (which went stale).
Original implementation by @hemanthsavasere and this PR rebases on current main, addresses review feedback and adds the follow-ups listed below.
  
  Closes #388
  Closes #442
  
  ## Original work
  - `Datum::Row`, `InternalRow::get_row`, CompactedRow ROW deserializer
  - `NestedRowWriter`, ROW arm in compacted key encoder
  - ROW round-trip unit tests
  
  ## Added in this revival
  - `field_id` machinery (server requires unique ids across nested ROW —
    matches Java's `ReassignFieldId`)
  - `ARRAY<ROW>` support
  - Arrow log-path nested-ROW disambiguation
  - Per-row `Vec<OnceLock>` allocation killed
  - new tests
  
  left followup https://github.com/apache/fluss-rust/issues/543 to address performance with nested structures, as this is PR is big enough and the gap comes from pre-existing code

### DIFF
--- a/crates/fluss/src/row/column.rs
+++ b/crates/fluss/src/row/column.rs
@@ -17,11 +17,17 @@
 
 use crate::error::Error::IllegalArgument;
 use crate::error::Result;
-use crate::row::{GenericRow, InternalRow};
+use crate::metadata::{DataType, RowType};
+use crate::record::from_arrow_type;
+use crate::row::binary_array::FlussArrayWriter;
 use crate::row::datum::{Date, Datum, Time, TimestampLtz, TimestampNtz};
+use crate::row::{Decimal, FlussArray, GenericRow, InternalRow};
 use arrow::array::{
-    Array, AsArray, BinaryArray, BooleanArray, FixedSizeBinaryArray, ListArray, RecordBatch,
-    StringArray,
+    Array, AsArray, BinaryArray, BooleanArray, Date32Array, Decimal128Array, FixedSizeBinaryArray,
+    Float32Array, Float64Array, Int8Array, Int16Array, Int32Array, Int64Array, ListArray,
+    RecordBatch, StringArray, StructArray, Time32MillisecondArray, Time32SecondArray,
+    Time64MicrosecondArray, Time64NanosecondArray, TimestampMicrosecondArray,
+    TimestampMillisecondArray, TimestampNanosecondArray, TimestampSecondArray,
 };
 use arrow::datatypes::{
     DataType as ArrowDataType, Date32Type, Decimal128Type, Float32Type, Float64Type, Int8Type,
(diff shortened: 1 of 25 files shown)


fresh-borzoni:
@leekeiabstraction @charlesdong1991 @luoyuxia PTAL 🙏

charlesdong1991:
Nice! thanks for reviving the PR, will take a look tomorrow! 🙏

crates/fluss/src/row/column.rs
```diff
@@ -433,6 +627,40 @@ impl InternalRow for ColumnarRow {
...
+            .row_column_indices
+            .iter()
+            .position(|&i| i == pos)
+            .ok_or_else(|| IllegalArgument {
+                message: format!("get_row called on non-ROW column at position {pos}"),
+            })?;
+        let column = self.record_batch.column(pos);
+        // Children of a null parent may carry stale bytes; caller must
+        // check is_null_at first rather than rely on what we'd read.
+        if column.is_null(self.row_id) {
```
charlesdong1991, line 642: i wonder should user check `is_null_at` before `get_row`?

crates/fluss/src/row/column.rs
```diff
@@ -35,25 +41,71 @@ use std::sync::Arc;
...
 
+    pub fn fluss_row_type(&self) -> Option<&Arc<RowType>> {
+        self.fluss_row_type.as_ref()
+    }
+
     pub fn set_row_id(&mut self, row_id: usize) {
-        self.row_id = row_id
+        self.row_id = row_id;
+        for lock in self.row_caches.iter_mut() {
+            *lock = std::sync::OnceLock::new();
```
charlesdong1991, line 107: you mentioned in PR description that `Per-row Vec<OnceLock> allocation killed`, looking at this, i wonder if it is still the case that it will allocate OnceLock on every row iteration?

crates/fluss/src/row/field_getter.rs
```diff
@@ -183,8 +187,9 @@ impl InnerFieldGetter {
             InnerFieldGetter::TimestampLtz { pos, precision } => {
                 Datum::TimestampLtz(row.get_timestamp_ltz(*pos, *precision)?)
             }
-            // TODO: add Map and Row field getter support once their binary forms are implemented.
+            // TODO: add Map field getter support once its binary form is implemented.
             InnerFieldGetter::Array { pos } => Datum::Array(row.get_array(*pos)?),
+            InnerFieldGetter::Row { pos } => Datum::Row(Box::new(row.get_row(*pos)?.clone())),
```
charlesdong1991, line 192: i wonder how performant this will be since it seems it clones the whole generic row? maybe we can flag awareness on hot scan paths here and revisit if needed later because the PR is huge already

charlesdong1991 (commented):
Thanks for the great work, overall looks great!! 👍 Honestly i paid a bit more attention on the addition in this revival 😅 and only have a couple minor comments, given the size of PR, we can probably better to have follow-up PRs instead.

In follow-up PRs we can also add corresponding docs to reflect the new change!

crates/fluss/src/row/column.rs
fresh-borzoni: it will disappear after https://github.com/apache/fluss-rust/issues/543

crates/fluss/src/row/column.rs
fresh-borzoni: ditto

crates/fluss/src/row/field_getter.rs
fresh-borzoni: ditto

fresh-borzoni:
Ty for the review @charlesdong1991 
Yes, I have followup planned, where I plan to switch to typed writers, so I will include docs there as well

PR #559, opened 2026-05-23
[test] Add Rust integration coverage for RecordBatchLogReader bounded reads.

slfan1989:
### Purpose

Linked issue: close #558 

Add Rust-level integration coverage for `RecordBatchLogReader` bounded reads.

`RecordBatchLogReader` already has unit coverage for batch filtering and Python binding coverage for `to_arrow_batch_reader` guard/drop behavior, but it lacks Rust end-to-end integration coverage for:

- `new_until_offsets` stopping semantics
- `new_until_latest` with partitioned log tables

This PR adds those missing integration tests.

### Brief change log

- Added a Rust integration test for `RecordBatchLogReader::new_until_offsets`.
  - Creates a log table.
  - Appends records.
  - Subscribes from a non-zero offset.
  - Reads with an explicit stopping offset.
  - Verifies records at or beyond the stop offset are not returned.

- Added a Rust integration test for `RecordBatchLogReader::new_until_latest` on partitioned tables.
  - Creates a partitioned log table.
  - Adds `US` and `EU` partitions.
  - Appends records to both partitions.
  - Subscribes partition buckets.
  - Verifies all records present at reader creation are returned.

- Updated the comment in `RecordBatchLogReader` to point to the new Rust integration coverage.

### Tests

Verified locally.

### API and Format

No API or storage format changes.

### Documentation

No user-facing documentation changes.

### DIFF
--- a/crates/fluss/tests/integration/record_batch_log_reader.rs
+++ b/crates/fluss/tests/integration/record_batch_log_reader.rs
@@ -1,0 +1,528 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#[cfg(test)]
+mod reader_test {
+    use crate::integration::utils::{
+        create_partitions, create_table, extract_ids_from_batches, get_shared_cluster,
+    };
+    use arrow::array::record_batch;
+    use fluss::client::{EARLIEST_OFFSET, FlussConnection, RecordBatchLogReader};
+    use fluss::config::{Config, NoKeyAssigner};
+    use fluss::metadata::{DataTypes, Schema, TableBucket, TableDescriptor, TablePath};
+    use fluss::rpc::message::OffsetSpec;
+    use std::collections::HashMap;
+    use std::time::Duration;
+
+    #[tokio::test]
+    async fn until_offsets_stops_at_explicit_offset() {
+        let cluster = get_shared_cluster();
+        let connection = cluster.get_fluss_connection().await;
+        let admin = connection.get_admin().expect("Failed to get admin");
+
+        let table_path = TablePath::new("fluss", "test_reader_until_offsets");
+        let table_descriptor = TableDescriptor::builder()
+            .schema(
+                Schema::builder()
+                    .column("id", DataTypes::int())
+                    .column("name", DataTypes::string())
+                    .build()
+                    .expect("Failed to build schema"),
+            )
+            .build()
+            .expect("Failed to build table");
+        create_table(&admin, &table_path, &table_descriptor).await;
+        tokio::time::sleep(Duration::from_secs(1)).await;
+
+        let table = connection
+            .get_table(&table_path)
+            .await
+            .expect("Failed to get table");
+        let writer = table
+            .new_append()
+            .expect("Failed to create append")
+            .create_writer()
+            .expect("Failed to create writer");
+        writer
+            .append_arrow_batch(
+                record_batch!(
+                    ("id", Int32, [1, 2, 3, 4, 5, 6]),
+                    ("name", Utf8, ["a", "b", "c", "d", "e", "f"])
+                )
+                .unwrap(),
+            )
+            .expect("Failed to append batch");
+        writer.flush().await.expect("Failed to flush");
+
+        let scanner = table
+            .new_scan()
+            .create_record_batch_log_scanner()
+            .expect("Failed to create record batch scanner");
+        scanner
+            .subscribe(0, 1)
+            .await
+            .expect("Failed to subscribe from offset 1");
+
+        let table_id = table.get_table_info().table_id;
+        let mut reader = RecordBatchLogReader::new_until_offsets(
+            scanner,
+            HashMap::from([(TableBucket::new(table_id, 0), 4)]),
+        )
+        .expect("Failed to create record batch reader");
+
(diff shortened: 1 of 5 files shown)


slfan1989:
@fresh-borzoni @leekeiabstraction Could you please help review this PR when you have time?

fresh-borzoni (commented):
@slfan1989 Ty for the PR, great addition, LGTM overall, couple of minor comments 👍 

Let's move this to a separate file for testing RecordBatchLogReader feature and then add some additional scenarios:
*until_offsets_with_empty_range (stop == start ->  no batches)
* until_offsets_past_end_of_log (graceful finish past actual end)
* until_offsets_multi_bucket (multiple buckets in the HashMap)

also `extract_ids_from_batches` overlaps with `extract_ids` in test_project test, mb we wish to factor this out to utils and reuse.

fresh-borzoni:
@charlesdong1991 Can you also take a look, please? 
Since this was initially your feature to propose :)

slfan1989:
> @slfan1989 Ty for the PR, great addition, LGTM overall, couple of minor comments 👍
> 
> Let's move this to a separate file for testing RecordBatchLogReader feature and then add some additional scenarios: *until_offsets_with_empty_range (stop == start -> no batches)
> 
> * until_offsets_past_end_of_log (graceful finish past actual end)
> * until_offsets_multi_bucket (multiple buckets in the HashMap)
> 
> also `extract_ids_from_batches` overlaps with `extract_ids` in test_project test, mb we wish to factor this out to utils and reuse.

@fresh-borzoni Thank you for the review and suggestions!

Addressed most of the comments:

- Moved the `RecordBatchLogReader` integration coverage into a dedicated test file: `record_batch_log_reader.rs`.
- Added `until_offsets_with_empty_range`.
- Added `until_offsets_multi_bucket`.
- Moved `extract_ids_from_batches` into `integration/utils.rs` and reused it from both `log_table.rs` and `record_batch_log_reader.rs`.

I left `until_offsets_past_end_of_log` out for now because it requires a behavior/API decision. 

The current `new_until_offsets` implementation waits until the requested stop offset is observed. If the stop offset is beyond the current log end, it keeps polling for future records rather than finishing. 

Supporting graceful completion would require snapshot-style semantics by querying latest offsets at reader creation, which needs `FlussAdmin` and async work. 

I think that should be handled separately from this test-only PR.

Thanks again!

crates/fluss/tests/integration/record_batch_log_reader.rs
```diff
@@ -0,0 +1,368 @@
...
+        assert_eq!(ids, vec![1, 2, 3, 4, 5, 6, 7, 8]);
+
+        admin
+            .drop_table(&table_path, false)
+            .await
+            .expect("Failed to drop table");
+    }
+
+    #[tokio::test]
+    async fn until_latest_reads_partitioned_table() {
```
charlesdong1991, line 435: query_latest_offsets should be applied on both partitioned and non-partitioned tables, can we check if non-partition one has integration test? and if not, add it?

crates/fluss/tests/integration/record_batch_log_reader.rs
```diff
@@ -0,0 +1,368 @@
...
+            .new_scan()
+            .create_record_batch_log_scanner()
+            .expect("Failed to create record batch scanner");
+        for partition in admin
+            .list_partition_infos(&table_path)
+            .await
+            .expect("Failed to list partition infos")
+        {
+            scanner
+                .subscribe_partition(partition.get_partition_id(), 0, EARLIEST_OFFSET)
```
charlesdong1991, line 502: i assume we use 0 here because default bucket number if 1, can you add an inline comment here for future reference if we want to have multiple buckets?

crates/fluss/tests/integration/record_batch_log_reader.rs
```diff
@@ -0,0 +1,368 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#[cfg(test)]
+mod reader_test {
```
charlesdong1991, line 21: > until_offsets_past_end_of_log (graceful finish past actual end)

i think what @fresh-borzoni meant might be the scenario where stop_at past current end, and future records arrive and cross it, which i think doesn't need API change, basically to cover wait and then resume behaviour of the loop. 

API change probably is needed if we want immediate graceful completion against arbtrary offsets

charlesdong1991 (commented):
thanks for the PR, overall very nice, just minor comments!

crates/fluss/tests/integration/record_batch_log_reader.rs
slfan1989: Thanks for pointing this out!

Added `until_latest_reads_non_partitioned_table`, so `new_until_latest` now has integration coverage for both non-partitioned and partitioned tables.

crates/fluss/tests/integration/record_batch_log_reader.rs
slfan1989: Added an inline comment explaining that bucket `0` is used because the table uses the default single-bucket layout, and that future multi-bucket coverage should subscribe all buckets per partition.

fresh-borzoni:
@slfan1989 Ty for the changes, LGTM
Can you rebase and resolve conflicts?

crates/fluss/tests/integration/record_batch_log_reader.rs
fresh-borzoni: yes, I should have been more specific. Sorry for the confusion

slfan1989: Got it, that makes sense. I interpreted `past_end_of_log` as immediate completion when the requested stop offset is beyond the current log end, which would require querying/clamping latest offsets.

I'll add a test for the wait-and-resume scenario instead: create a reader with a stop offset past the current end, start collecting, append more records later so the log crosses the stop offset, and verify the reader resumes and finishes.

slfan1989:
> @slfan1989 Ty for the changes, LGTM
> Can you rebase and resolve conflicts?

@fresh-borzoni  Thanks for the review!

I have rebased the branch on the latest `main` and resolved the conflicts. The updated changes have been pushed, and I’m waiting for CI to finish now.

fresh-borzoni (approved):
@slfan1989 LGTM 👍 

@charlesdong1991 Do you have additional comments?

charlesdong1991 (approved):
thanks for the changes, very nice! 👍

fresh-borzoni:
@slfan1989 Thank you for the contribution, merged 👍

PR #560, opened 2026-05-23
[rust] integration tests for MAP dt + restructure ITs for complex dts

fresh-borzoni:
closes #549

Added Map integration tests, consolidated dt tests to avoid overhead in table creations and exponential tests per complex dts.

### DIFF
--- a/crates/fluss/src/row/column.rs
+++ b/crates/fluss/src/row/column.rs
@@ -661,24 +661,18 @@ impl InternalRow for ColumnarRow {
         };
 
         let column = self.column(pos)?;
-        let element_field = match column.data_type() {
-            ArrowDataType::List(field) => field,
+        match column.data_type() {
+            ArrowDataType::List(_) => {}
             other => {
                 return Err(IllegalArgument {
                     message: format!("expected List array at position {pos}, got {other:?}"),
                 });
             }
-        };
-
-        let actual_element_type = from_arrow_field(element_field)?;
-        if actual_element_type != *element_fluss_type {
-            return Err(IllegalArgument {
-                message: format!(
-                    "Arrow list element type {:?} does not match expected Fluss type {:?}",
-                    actual_element_type, element_fluss_type
-                ),
-            });
         }
+
+        // `to_arrow_type` is lossy (e.g. TIMESTAMP_LTZ → plain Arrow Timestamp);
+        // trust the Fluss schema and let the per-element conversion below catch
+        // real shape mismatches.
 
         let list_arr = column
             .as_any()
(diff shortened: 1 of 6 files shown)


fresh-borzoni:
@charlesdong1991 @leekeiabstraction @luoyuxia  PTAL 🙏 , I need this as a preparation to refactor complex types to use typed writers, it's mostly tests consolidation and new map tests.

Appreciate if you can take a quick look

crates/fluss/src/row/column_writer.rs
```diff
@@ -940,6 +939,57 @@ fn write_map_into(
     Ok(())
 }
 
+// FlussArray carries no schema; nested row/map elements need the typed
```
fresh-borzoni, line 942: will be fixed with refactoring, now I need this for test to pass

fresh-borzoni:
@luoyuxia addressed Copilot comments, added couple of helpers to help with boundaries and shared schema between log and kv for "all datatypes" tests, extended kv tests to include more precisions(it was required to unify).
PTAL 🙏

crates/fluss/src/row/column.rs
```diff
@@ -661,25 +661,19 @@ impl InternalRow for ColumnarRow {
...
-        if actual_element_type != *element_fluss_type {
-            return Err(IllegalArgument {
-                message: format!(
-                    "Arrow list element type {:?} does not match expected Fluss type {:?}",
-                    actual_element_type, element_fluss_type
-                ),
-            });
         }
 
+        // `to_arrow_type` is lossy (e.g. TIMESTAMP_LTZ → plain Arrow Timestamp);
```
charlesdong1991, line 673: so to confirm since not mentioned in PR description: i guess this will be a behaviour, correct? that users won't know schema mismatch when calling get_array on position

crates/fluss/tests/integration/utils.rs
```diff
@@ -176,3 +178,200 @@ pub async fn create_partitions(
...
+            DataTypes::timestamp_ltz_with_precision(6),
+        ),
+        (
+            "col_ts_ltz_ns_neg",
+            DataTypes::timestamp_ltz_with_precision(9),
+        ),
+    ]
+}
+
+#[derive(Default)]
```
charlesdong1991, line 303: nice idea on ColumnPlan that helps reduce redundancy a lot!

Just a nit: do we need `derive[Default]` for internal test only feature?

crates/fluss/tests/integration/kv_table.rs
```diff
@@ -2175,4 +1251,522 @@ mod kv_table_test {
             .await
             .expect("Failed to drop table");
     }
+
+    /// KV upsert + lookup against a schema covering every supported data type.
+    #[tokio::test]
+    async fn all_supported_datatypes() {
```
charlesdong1991, line 1257: Very nit: somehow i am a bit conservative to have such huge test, i think now if something breaks, developers will see all unrelated assertions and not sure if there is meaningful tie with the failing column index? We can revisit in future if encounter such issue in development

fresh-borzoni:
Merging to continue working on typed writers refactoring

charlesdong1991 (commented):
very minor comments, no blocker for me

crates/fluss/tests/integration/kv_table.rs
fresh-borzoni: Each row's assertions include the column name in their failure messages, so a break should at least point to the offending column.

but we can revisit later if it becomes a concern 👍

crates/fluss/src/row/column.rs
fresh-borzoni: Intentional. The old strict check tripped on lossy Arrow round-trips (e.g. TIMESTAMP_LTZ -> plain Arrow Timestamp) and rejected valid schemas. 

Real shape mismatches still error from the per-element conversion below

crates/fluss/tests/integration/utils.rs
fresh-borzoni: well, it was easier this way, another option - inline Self { cols: vec![], index: HashMap::new(), sections: vec![] }, but it's all the same. Arguably derive is a bit cleaner overall

PR #566, opened 2026-05-26
 [elixir] feat: Extend Fluss.Config with writer tuning options

nicolazar:
### Purpose

Linked issue: closes #567, part of #463.

### Brief change log

This PR extends the Elixir bindings with new config options for writer tuning, such as: writer acks, retries, request max size, idempotence, buffer memory / wait timeout, max inflight requests per bucket, bucket assigner enum.

We first extend the NifConfig on the rust side to expose the new options, then we add them to `Fluss.Config`, with the corresponding setters. The upstream defaults are preserved if fields are not set, but callers may set overrides per field.

Validation continues to be handled by `FlussConnection::new` ([`crates/fluss/src/client/connection.rs:40-48`](https://github.com/apache/fluss-rust/blob/main/crates/fluss/src/client/connection.rs#L40-L48)) which already calls `validate_writer` before any TCP I/O.

### Tests

Tests for the config setters are added in the new `test/config_tests.exs`: 

  - `new/1` defaults — all non-`bootstrap_servers` fields default to `nil`.
  - One per-setter test for each of the 8 new setters.
  - `set_writer_bucket_no_key_assigner/2`: happy path (`:sticky`) and guard failure (`assert_raise FunctionClauseError` for unsupported atoms).
  - Chained-setters test using `|>`.

Issue #463, opened 2026-03-29
[elixir] Extended configuration options


fresh-borzoni:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/fluss-rust/issues) and found nothing similar.

### Description

depends on https://github.com/apache/fluss-rust/pull/452

Add writer tuning (acks, retries, idempotence, buffer size, inflight limit), scanner tuning (prefetch, concurrency, poll records, fetch bytes), security/SASL config, connect timeout. 

### Willingness to contribute

- [ ] I'm willing to submit a PR!

PR #574, opened 2026-05-27
[elixir] feat: Extend Fluss.Config with scanner tuning options

nicolazar:
### Purpose

Linked issue: close #568, part of #463  

### Brief change log

This PR follows up on #566 to extend Fluss.Config - this time with 8 more scanner tuning options, such as remote log prefetch, download concurrency, max poll records, fetch size/time limits.

We follow the same approach as in the linked PR.

### Tests

Additional tests cover the newly added options, one per field.

### DIFF
--- a/bindings/elixir/lib/fluss/config.ex
+++ b/bindings/elixir/lib/fluss/config.ex
@@ -33,6 +33,14 @@ defmodule Fluss.Config do
 
   @enforce_keys [:bootstrap_servers]
   defstruct bootstrap_servers: nil,
+            remote_file_download_thread_num: nil,
+            scanner_log_fetch_max_bytes: nil,
+            scanner_log_fetch_max_bytes_for_bucket: nil,
+            scanner_log_fetch_min_bytes: nil,
+            scanner_log_fetch_wait_max_time_ms: nil,
+            scanner_log_max_poll_records: nil,
+            scanner_remote_log_prefetch_num: nil,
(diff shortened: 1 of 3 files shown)


bindings/elixir/lib/fluss/config.ex
```diff
@@ -74,6 +90,44 @@ defmodule Fluss.Config do
   def set_bootstrap_servers(%__MODULE__{} = config, servers) when is_binary(servers),
     do: %{config | bootstrap_servers: servers}
 
+  @spec set_remote_file_download_thread_num(t(), non_neg_integer()) :: t()
+  def set_remote_file_download_thread_num(%__MODULE__{} = config, threads)
+      when is_integer(threads),
```
fresh-borzoni, line 95: nit: do you mind to tighten the guards to match type specs we use? 
here and in other setters through config

fresh-borzoni (commented):
@nicolazar Ty for the PR, LGTM overall, one comment

bindings/elixir/lib/fluss/config.ex
nicolazar: @fresh-borzoni - I can do that, but in my opinion might be better to have a separate PR (and issue) where we add dialyzer and fix all the warnings that would appear. What do you think?

fresh-borzoni: Sure, do you mind to file an issue for this?

PR #576, opened 2026-05-28
[ci] Fix rustdoc warning and promote warning to error in CI

charlesdong1991:
### Purpose

it seems we started to have rustdoc warnings, i think we should add CI enforcement to precent documentation regression

fixed those 2 warnings in PR

### DIFF
--- a/crates/fluss/src/record/kv/value_record_batch.rs
+++ b/crates/fluss/src/record/kv/value_record_batch.rs
@@ -75,7 +75,7 @@ impl ValueRecordBatch {
     }
 
     /// Returns one byte range per record, each spanning `[SchemaId | Value]`:
-    /// the payload [`crate::row::FixedSchemaDecoder::decode`] expects. Index
+    /// the payload that `FixedSchemaDecoder::decode` expects. Index
     /// [`Self::data`] with a returned range to get it without copying.
     pub fn value_ranges(&self) -> Result<Vec<Range<usize>>> {
         let count = self.record_count()?;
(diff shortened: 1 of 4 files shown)


charlesdong1991:
cc @fresh-borzoni @leekeiabstraction let me know what you think of enforcement 🙏

crates/fluss/src/record/kv/value_record_batch.rs
```diff
@@ -75,7 +75,7 @@ impl ValueRecordBatch {
     }
 
     /// Returns one byte range per record, each spanning `[SchemaId | Value]`:
-    /// the payload [`crate::row::FixedSchemaDecoder::decode`] expects. Index
+    /// the payload that `FixedSchemaDecoder::decode` expects. Index
```
leekeiabstraction: This removes the linking, what was the error and could it be resolved while retaining the link?

leekeiabstraction (commented):
TY for the PR, left a question

crates/fluss/src/record/kv/value_record_batch.rs
charlesdong1991: so warning is:
```
error: public documentation for `value_ranges` links to private item `crate::row::FixedSchemaDecoder::decode`
```

so IMHO linting is correct, link points to something that doesn't exist in public doc output. we can fix by making this `FixedSchemaDecoder` pub (which i feel unless is intentional, maybe not to expand public api), or allow private intra doc links for this value_ranges... 

not sure if either is better, i went with not touching codebase, but happy to pick any option you prefer as i don't have strong preference

leekeiabstraction: Gotcha. If I understand correctly, these (ValueRecordBatch and FixedSchemaDecoder) are mainly internals of fluss-rust. Agree with removing the link as valuable this is not as valuable in internal functions. I might go one step further and make these `pub (crate)`.

charlesdong1991: > make these pub (crate).

oh yeah, this will also work!

charlesdong1991: done, thanks for review! @leekeiabstraction 🙏

leekeiabstraction (approved):
LGTM. TY for your contribution!

PR #577, opened 2026-05-28
[elixir] feat: Extend Fluss.Config with security, SASL and connect timeout options

nicolazar:
### Purpose

Linked issue: close #569 and the parent #463. 

### Brief change log

This PR follows up on #574 to extend Fluss.Config with configuration options for security protocol, SASL and connect timeout. To ensure that the SASL password, if set, gets redacted in logs, we customise how Fluss.Config gets inspected using the [Inspect protocol](https://hexdocs.pm/elixir/Inspect.html); it would appear as `[REDACTED]` in logs.

We otherwise follow the same approach as in the related PRs.

### Tests

New tests cover the setters that were added for these options, one for each function.

### DIFF
--- a/bindings/elixir/lib/fluss/config.ex
+++ b/bindings/elixir/lib/fluss/config.ex
@@ -33,6 +33,7 @@ defmodule Fluss.Config do
 
   @enforce_keys [:bootstrap_servers]
   defstruct bootstrap_servers: nil,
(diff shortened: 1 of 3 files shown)


fresh-borzoni (approved):
@nicolazar Ty, LGTM 👍

PR #578, opened 2026-05-29
[rust] typed column-vector dispatch

fresh-borzoni:
### Summary

closes https://github.com/apache/fluss-rust/issues/543

ColumnarRow now holds typed Arrow column vectors built once per batch instead of downcasting on every read. Nested ARRAY/MAP/ROW return lazy slice views - no copy unless the caller asks for a binary form via try_into_binary(). 

Mirrors Java's VectorizedColumnBatch.

### DIFF
--- a/crates/fluss/src/record/arrow.rs
+++ b/crates/fluss/src/record/arrow.rs
@@ -22,8 +22,9 @@ use crate::compression::{
 use crate::error::{Error, Result};
 use crate::metadata::{DataField, DataType, RowType};
 use crate::record::{ChangeType, ScanRecord};
+use crate::row::column_vector::TypedBatch;
 use crate::row::column_writer::{ColumnWriter, round_up_to_8};
-use crate::row::{ColumnarRow, InternalRow, arrow_row_column_indices, fluss_row_column_indices};
+use crate::row::{ColumnarRow, InternalRow};
 use arrow::array::{ArrayBuilder, ArrayRef};
 use arrow::{
     array::RecordBatch,
@@ -1667,21 +1668,12 @@ impl Iterator for ArrowLogRecordIterator {
 }
 
 pub struct ArrowReader {
-    record_batch: Arc<RecordBatch>,
-    row_type: Arc<RowType>,
-    fluss_row_type: Option<Arc<RowType>>,
-    row_column_indices: Arc<[usize]>,
+    batch: Arc<TypedBatch>,
 }
 
 impl ArrowReader {
     pub fn new(record_batch: Arc<RecordBatch>, row_type: Arc<RowType>) -> Self {
-        let row_column_indices = arrow_row_column_indices(&record_batch);
-        ArrowReader {
-            record_batch,
-            row_type,
-            fluss_row_type: None,
-            row_column_indices,
-        }
+        Self::new_with_fluss_row_type(record_batch, row_type, None)
     }
 
     pub fn new_with_fluss_row_type(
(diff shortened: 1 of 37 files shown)


fresh-borzoni:
@charlesdong1991 @leekeiabstraction PTAL 🙏

crates/fluss/src/row/column.rs
```diff
@@ -15,1197 +15,246 @@
...
-        for lock in self.row_caches.iter_mut() {
-            *lock = std::sync::OnceLock::new();
-        }
     }
 
     pub fn get_row_id(&self) -> usize {
         self.row_id
     }
 
+    /// Returns the source `RecordBatch`. Panics on a nested cursor (returned
```
leekeiabstraction: Curious on why panic as opposed to Error? Both ColumnarRow and this function are pub. IMO we should reduce foot guns if possible.

leekeiabstraction (commented):
Ty for the PR. Haven't managed to read it all but left a question. Will pick it up again tomorrow

crates/fluss/src/record/arrow.rs
```diff
@@ -1667,52 +1668,33 @@ impl Iterator for ArrowLogRecordIterator {
...
         record_batch: Arc<RecordBatch>,
         row_type: Arc<RowType>,
         fluss_row_type: Option<Arc<RowType>>,
     ) -> Self {
-        let row_column_indices = match &fluss_row_type {
-            Some(rt) => fluss_row_column_indices(rt),
-            None => arrow_row_column_indices(&record_batch),
-        };
+        let schema = fluss_row_type.as_deref().unwrap_or(&row_type);
+        let typed = TypedBatch::build(&record_batch, schema)
```
charlesdong1991: so now build is fallible, and a malformed batch panics the scan instead of surfacing error?

crates/fluss/src/row/column_vector.rs
```diff
@@ -0,0 +1,650 @@
...
+            Self::Map(a, _, _) => a,
+            Self::Row(a, _) => a,
+        }
+    }
+
+    pub(crate) fn is_null_at(&self, row_id: usize) -> bool {
+        self.outer_array().is_null(row_id)
+    }
+
+    pub(crate) fn get_boolean(&self, row_id: usize) -> Result<bool> {
```
charlesdong1991, line 231: i wonder now since all is centralized in TypedColumn, probably it is cheap to bound check for  top level getters?

crates/fluss/src/row/columnar.rs
```diff
@@ -0,0 +1,364 @@
...
+    }
+
+    fn get_binary(&self, pos: usize, _length: usize) -> Result<&[u8]> {
+        self.element.get_binary(self.absolute_index(pos)?)
+    }
+    fn get_bytes(&self, pos: usize) -> Result<&[u8]> {
+        self.element.get_bytes(self.absolute_index(pos)?)
+    }
+
+    fn get_array(&self, pos: usize) -> Result<ArrayView<'_>> {
```
charlesdong1991, line 133: this is public right? should we update api-reference since the returned type has changed

crates/fluss/src/row/columnar.rs
```diff
@@ -0,0 +1,364 @@
...
+                    end - start,
+                )))
+            }
+            other => Err(IllegalArgument {
+                message: format!("expected ARRAY element at position {pos}, got {other:?}"),
+            }),
+        }
+    }
+
+    fn get_map(&self, pos: usize) -> Result<MapView<'_>> {
```
charlesdong1991, line 151: similar here

charlesdong1991 (commented):
thanks for great PR, i will need some more time to read it through and learn a bit better about current behaviour in java, just a quick look, will take another look later 🙏

leekeiabstraction (commented):
No further comments that what I've left earlier and also what was mentioned by @charlesdong1991 

As usual, great PR! Really like some of the refactor within this.

crates/fluss/src/row/column.rs
fresh-borzoni: Changed it to return Option<&RecordBatch> - nested rows just get None now instead of panicking, it's better - a good call.

crates/fluss/src/record/arrow.rs
fresh-borzoni: Yeah, agreed - it shouldn't panic on a bad batch off the wire, fixed

fresh-borzoni:
@leekeiabstraction @charlesdong1991 Ty for the review, addressed comments, PTAL 🙏

leekeiabstraction (approved):
LGTM, thank you for the PR!

charlesdong1991 (approved):
thanks for the great PR, very nice optimization! 👍

PR #581, opened 2026-05-30
[metrics] Add last_poll_seconds_ago scanner metric

charlesdong1991:
### Purpose

Add `lastPollSecondsAgo` consumer-liveness gauge to fluss-rust, this is the primary signal for detecting stuck consumers, and it must keep advancing between poll events even when no new poll happen.

Note: `metrics 0.24` facade has no scrape-time gauge supplier abstraction, so the value is pushed by a per-scanner 1-second background tokio task.

Linked issue: close #554 

### Tests

All tests passed.

### DIFF
--- a/crates/fluss/src/metrics.rs
+++ b/crates/fluss/src/metrics.rs
@@ -88,8 +88,7 @@ pub const SCANNER_POLL_IDLE_RATIO: &str = "fluss.client.scanner.poll_idle_ratio"
 /// skipped until the first `poll()` happens; Java's equivalent
 /// `lastPollSecondsAgo` returns roughly the current Unix-epoch seconds
 /// before the first poll (an unguarded `(now - 0)/1000`), which would trip
-/// every consumer-liveness alert on startup. Classified as `can-differ` per
-/// `java-rust-parity-criteria.mdc` since this is purely client-side display.
+/// every consumer-liveness alert on startup.
 ///
 pub const SCANNER_LAST_POLL_SECONDS_AGO: &str = "fluss.client.scanner.last_poll_seconds_ago";
 
--- a/crates/fluss/src/metrics.rs
+++ b/crates/fluss/src/metrics.rs
@@ -60,16 +60,15 @@ pub const CLIENT_REQUESTS_IN_FLIGHT: &str = "fluss.client.requests_in_flight";
 // Java reference: ScannerMetricGroup.java, LogScannerImpl.java
 //
 // These track consumer liveness and processing efficiency at the `poll()`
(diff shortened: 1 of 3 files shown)


charlesdong1991:
@fresh-borzoni @leekeiabstraction PTAL 🙏

crates/fluss/src/metrics.rs
```diff
@@ -160,6 +171,7 @@ impl ScannerMetrics {
         Self {
             time_between_poll_ms: scanner_gauge(SCANNER_TIME_BETWEEN_POLL_MS, database, table),
             poll_idle_ratio: scanner_gauge(SCANNER_POLL_IDLE_RATIO, database, table),
+            last_poll_seconds_ago: scanner_gauge(SCANNER_LAST_POLL_SECONDS_AGO, database, table),
```
leekeiabstraction, line 174: `scanner_gauge` is keyed by database and table. What if there are two scanners (with different projections for example) running and one of them is stuck? Is there a risk that metric of the healthy one overwrites/gets emitted over the value of the stuck one and consequently alert is not raised?

leekeiabstraction (commented):
TY for the PR, left a comment.

crates/fluss/src/metrics.rs
charlesdong1991: good call, it doesn't cover per-scanner dimension. I also checked on Java side, seems also not there, and actually i think it's best to create an issue in fluss and seek for feedback and implement there first, and then match here. 

wdyt? happy to create issue and contribute there first if you agree with assessment 🙏 thanks @leekeiabstraction

fresh-borzoni: @charlesdong1991 @leekeiabstraction - yes, I think it's the correct sequence

fresh-borzoni (approved):
@charlesdong1991 Ty for the PR, LGTM, though some comments are a bit hairy to my taste, not a blocker.

leekeiabstraction (approved):

crates/fluss/src/metrics.rs
fresh-borzoni: @charlesdong1991 But we still wait for the issue link here :)

charlesdong1991: Created: https://github.com/apache/fluss/issues/3415  🙏 let's see if anyone replies, otherwise, can try to come up with a PR to collect feedback. @fresh-borzoni

fresh-borzoni: Thank you 😊

PR #582, opened 2026-05-30
[elixir] chore: Add dialyzer and tighten Fluss.Config setter guards

nicolazar:
### Purpose

Linked issue: close #575 

### Brief change log

This PR adds [dialyzer](https://hex.pm/packages/dialyxir) and enforces it to pass in CI. `:dialyxir` is added as a dev/test dep with PLT pinned to `priv/plts/dialyzer.plt` (so the CI cache key is predictable).

We also add a new `mix check` alias to encapsulate all static checks in a single command for local verification.

Finally, we tighten `Fluss.Config` numeric setter guards: they declared `non_neg_integer()` typespecs, but guarded with `is_integer/1` — accepting negative values at runtime despite the spec.


fresh-borzoni (approved):
@nicolazar Ty for the PR, LGTM 👍

PR #601, opened 2026-06-06
[python] map and row complex types support

### LINKED ISSUE #602: [python] Add Map data type support
### Search before asking

### Description

Add Map dt to python bindings, restructure tests to match rusts

### Willingness to contribute

### LINKED ISSUE #603: [python] Add Row data type
### Search before asking

### Description

Add Row data type to python bindings, restructure tests to match rusts

### Willingness to contribute

fresh-borzoni:
closes https://github.com/apache/fluss-rust/issues/602
closes https://github.com/apache/fluss-rust/issues/603

Adds MAP and ROW support to the Python client, so nested columns work the same way ARRAY already does. A MAP reads back as a list of (key, value) tuples and a ROW as a dict keyed by field name, matching what pyarrow's to_pylist() produces and both can be nested arbitrarily and written from plain Python dicts. It works on both the per-row path and the zero-copy Arrow path.

Two core fixes fell out of writing the tests. Reading a timestamp-valued map on the lookup path was failing because FlussArray's size calculation misread non-compact timestamps, and TIMESTAMP_LTZ now keeps its UTC zone when converted to Arrow, so it round-trips consistently instead of coming back naive. Neither case was covered by the existing Rust integration tests.

Row writes from a dict are now nullability-aware: a nullable column may be omitted (it defaults to null), but a non-null or primary-key column has to be present. 

Tests reworked to match rust shape for data types supported

### DIFF
--- a/bindings/python/src/utils.rs
+++ b/bindings/python/src/utils.rs
@@ -18,6 +18,7 @@
 use crate::*;
 use arrow_pyarrow::{FromPyArrow, ToPyArrow};
 use arrow_schema::SchemaRef;
+use fcore::record::from_arrow_field;
 use std::sync::Arc;
 
 /// Utilities for schema conversion between PyArrow, Arrow, and Fluss
@@ -42,7 +43,7 @@ impl Utils {
     pub fn arrow_field_to_fluss_type(
         field: &arrow::datatypes::Field,
     ) -> PyResult<fcore::metadata::DataType> {
-        fcore::record::from_arrow_field(field).map_err(|e| FlussError::from_core_error(&e))
+        from_arrow_field(field).map_err(|e| FlussError::from_core_error(&e))
     }
 
     /// Convert Fluss DataType to string representation, appending " NOT NULL"
--- a/bindings/python/src/utils.rs
+++ b/bindings/python/src/utils.rs
@@ -36,82 +36,13 @@ impl Utils {
         })
     }
 
-    /// Convert an Arrow Field to a Fluss DataType, preserving nullability.
+    /// Convert an Arrow Field to a Fluss DataType. Delegates to core's canonical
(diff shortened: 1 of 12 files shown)


fresh-borzoni:

bindings/python/src/utils.rs
```diff
@@ -36,82 +36,13 @@ impl Utils {
...
-                )));
-            }
-        };
-
-        if field.is_nullable() {
-            Ok(fluss_type)
-        } else {
-            Ok(fluss_type.as_non_nullable())
-        }
+        fcore::record::from_arrow_field(field).map_err(|e| FlussError::from_core_error(&e))
```
charlesdong1991: just to confirm: it seems the match branch here handles more types than core's from_array_field, is it intentional?

charlesdong1991 (commented):
very nice PR as always! Just a minor comment, PTAL 👍 

In addition, a small nit: does example.py need to be updated now? i have a feel that it needs... 

also i think example.py is not guaranteed to be executable, maybe we can create a separate issue to make such guarantee as i find it important for users, happy to work on that ^^

bindings/python/src/utils.rs
fresh-borzoni: My bad, I didn't notice that we lack some of the pyArrow supported types, thank you for flagging

fresh-borzoni:
@charlesdong1991 Added additional types to delegated call, PTAL 🙏 

Agreed it'd be nice to showcase nested types, it's actually flat today (doesn't even use ARRAY).
We should also run examples in CI, so that we don't break them and make sure it works
Appreciate if you can help with this 👍

charlesdong1991 (approved):
thanks for the PR, very nice! 👍

charlesdong1991:
i logged an issue here for reference: https://github.com/apache/fluss-rust/issues/607 @fresh-borzoni

PR #606, opened 2026-06-06
[c++] add prefix lookup

### LINKED ISSUE #537: [c++] Support prefix lookup
### Search before asking

### Description

The CXX bridge needs three new entries - the opaque PrefixKeyLookuper type, a constructor that takes the column-names list, and the lookup call itself  and on the C++ side a class mirroring the existing Lookuper's raw-pointer + Destroy() RAII pattern (move-only, manual cleanup). 

The one thing that doesn't have a perfect precedent is the multi-row result: the existing LookupResult only models a single row, so a new small PrefixLookupResult is needed with Size() and GetRow(i) returning a RowView. That's not a new design either - ScanResult already does multi-row over the same zero-copy row machinery, so it's mostly copying that shape into a smaller class. 

Add an example program, a couple of integration tests, and a docs entry

### Willingness to contribute

fresh-borzoni:
closes https://github.com/apache/fluss-rust/issues/537

Adds bucket-key prefix lookup to the C++ binding, which the Rust and Python clients already had. You create a lookuper with table.NewPrefixLookup({cols}, out) and call PrefixLookup(prefix_row, result) and it returns zero or more rows whose primary key starts with the given prefix, read through the same zero-copy row API as point lookups.

 It's a single-step factory rather than a builder, matching how NewLookup is collapsed elsewhere in the C++ surface.
 
 I leave possible consolidation of plv and lv to the followup.

### DIFF
--- a/bindings/cpp/src/table.cpp
+++ b/bindings/cpp/src/table.cpp
@@ -1607,8 +1607,9 @@ Result PrefixLookuper::PrefixLookup(const GenericRow& prefix_row, PrefixLookupRe
                                  std::string(result_box->plv_error_message()));
     }
 
-    // Wrap the raw pointer immediately so it's never leaked on exception, then
-    // build the column map eagerly — shared by all PrefixRowViews.
+    // Take ownership of the FFI box first (~PrefixData calls from_raw), so the
+    // column-map loop below can't leak it if a string/map allocation throws.
+    // The map is built eagerly and shared by all PrefixRowViews.
     auto data = std::make_shared<detail::PrefixData>(result_box.into_raw(), detail::ColumnMap{});
     auto col_count = data->raw->plv_field_count();
     for (size_t i = 0; i < col_count; ++i) {
--- a/bindings/cpp/src/table.cpp
+++ b/bindings/cpp/src/table.cpp
@@ -431,6 +431,12 @@ detail::ScanData::~ScanData() {
     }
 }
 
+detail::PrefixData::~PrefixData() {
+    if (raw) {
+        rust::Box<ffi::PrefixLookupResultInner>::from_raw(raw);
+    }
+}
+
 // ============================================================================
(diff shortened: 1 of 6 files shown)


fresh-borzoni:

bindings/cpp/src/table.cpp
```diff
@@ -1405,6 +1561,65 @@ Result Lookuper::Lookup(const GenericRow& pk_row, LookupResult& out) {
...
+    }
+
+    auto result_box = lookuper_->prefix_lookup(*prefix_row.inner_);
+    if (result_box->plv_has_error()) {
+        return utils::make_error(result_box->plv_error_code(),
+                                 std::string(result_box->plv_error_message()));
+    }
+
+    // Wrap the raw pointer immediately so it's never leaked on exception, then
+    // build the column map eagerly — shared by all PrefixRowViews.
```
charlesdong1991: very very nit question for myself: is "never leaked" comment only referring to result_box.into_raw? or the whole line below?

charlesdong1991 (approved):
nice one!! 👍

bindings/cpp/src/table.cpp
fresh-borzoni: Good question 👍 . Think of it as a hand-off: the Rust Box owns the memory and frees it for you automatically. into_raw() gives that up and hands back a plain pointer and now nothing frees it unless we arrange it. So on the same step we hand it to PrefixData, whose destructor calls from_raw. That's why the "never leaked" is really about the loop after it: once PrefixData owns the pointer, if the map-building throws, it still gets cleaned up on the way out.

Honestly though, the bigger reason for the into_raw/from_raw  is just to keep the cxx Box type out of the public header and the exception-safety bit is more of a nice by-product, and a pretty standard "wrap it immediately" pattern.

charlesdong1991 (approved):

PR #609, opened 2026-06-07
[metrics] Add writer pipeline metrics

charlesdong1991:
### Purpose

Adds writer-pipeline observability metrics to rust, closing the writer-side gap for observability.

Linked issue: close #608

### Tests

All tests passed locally.

### API and Format

updated api-references

### DIFF
--- a/crates/fluss/src/client/write/sender.rs
+++ b/crates/fluss/src/client/write/sender.rs
@@ -747,7 +747,7 @@ impl Sender {
     /// batches. Invoked once at the start of `send_write_request`, before the
     /// leader lookup / connection / serialization steps, so every drained
     /// batch is counted exactly once per send attempt regardless of whether
-    /// the send later succeeds. Because this runs before serialization, 
+    /// the send later succeeds. Because this runs before serialization,
     /// `estimated_size_in_bytes` is the pre-serialization estimate rather than  
     /// the final encoded length.
     fn record_request_batch_metrics(&self, request_batches: &[ReadyWriteBatch]) {
--- a/crates/fluss/src/client/write/sender.rs
+++ b/crates/fluss/src/client/write/sender.rs
@@ -311,6 +311,15 @@ impl Sender {
         if batches.is_empty() {
             return Ok(());
         }
+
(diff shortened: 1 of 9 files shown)


charlesdong1991:
cc @fresh-borzoni @leekeiabstraction PTAL 🙏

crates/fluss/src/client/write/sender.rs
```diff
@@ -1058,6 +1108,56 @@ mod tests {
         Ok((batch, result_handle))
     }
 
+    fn build_cluster_arc_with_port(
```
fresh-borzoni: utils?

crates/fluss/src/client/write/sender.rs
```diff
@@ -371,8 +376,15 @@ impl Sender {
...
             };
 
-            // let's put in back into records_by_bucket
-            // since response handle will use it.
+            // Record attempted-send per-batch metrics after
+            // `build_write_request` (so `estimated_size_in_bytes` and
+            // `record_count` reflect the final serialized batch), before the
+            // actual request round-trip. Retries therefore contribute one
+            // sample per send attempt.
+            self.record_request_batch_metrics(&request_batches);
```
fresh-borzoni: If get_connection fails, we bail out before reaching record_request_batch_metrics, so those batches never get counted in records_send_total / bytes_send_total. 

Java still counts them since it updates metrics over the whole drained set regardless of send outcome. Arguably ours is even more accurate as we only count batches that actually got serialized and dispatched, but if we wish this, we need to change the semantics in Java first.

For now it's better to match.

fresh-borzoni (commented):
@charlesdong1991 Ty for the PR, LGTM overall, couple of comments, PTAL

crates/fluss/src/client/write/sender.rs
charlesdong1991: ok, let's then match java behaviour for now, thanks!

charlesdong1991:
@fresh-borzoni addressed comments, PTAL 🙏 thanks

fresh-borzoni (approved):
@charlesdong1991 Thank you, LGTM 👍

PR #611, opened 2026-06-07
[c++] Add MAP/ROW types, unify complex types under Value handle

fresh-borzoni:
closes https://github.com/apache/fluss-rust/issues/604
closes https://github.com/apache/fluss-rust/issues/605

Adds MAP and ROW column support to the C++ client, bringing it in line with the Rust core and Python.
  
The substantive change is on the read side. ARRAY reads went through a per-type accessor family hung off each result type, and extending that to MAP and ROW would have meant a per-type cross-product across three result containers (scan, lookup, prefix lookup). Instead, any complex column is now read through a single recursive Value handle: navigate with At / KeyAt / ValueAt / Field, read leaves with the Get* methods. That replaced the old ArrayView/MapView/RowValue types everywhere and nests to any depth.
  
For schema and writes, you declare nested columns with DataType::Map / DataType::Row and build values with the existing writers - no Arrow in user code. Schema::FromArrow stays as an escape hatch for callers that already have an arrow::Schema, internally the native types are lowered to Arrow when the table is created. 
The zero-copy scan path and the Arrow bulk path are left as-is.

* examples are left as a follow-up

### DIFF
--- a/bindings/cpp/src/admin.cpp
+++ b/bindings/cpp/src/admin.cpp
@@ -21,6 +21,8 @@
 #include "fluss.hpp"
 #include "lib.rs.h"
 #include "rust/cxx.h"
+#include "type_lowering.hpp"
+#include <arrow/c/bridge.h>
 #include <exception>
 
 namespace fluss {
@@ -58,8 +60,37 @@ Result Admin::CreateTable(const TablePath& table_path, const TableDescriptor& de
     }
 
     auto ffi_path = utils::to_ffi_table_path(table_path);
-    auto ffi_desc = utils::to_ffi_table_descriptor(descriptor);
 
+    // A MAP/ROW column can't go through the flat FFI encoding, so the schema is
+    // sent over Arrow instead (explicit via FromArrow, or lowered from native
+    // columns). Rust derives the columns from it, so the flat columns are dropped
+    // here; primary keys / metadata still come from the descriptor.
+    std::shared_ptr<arrow::Schema> arrow_schema = descriptor.schema.arrow_schema;
+    if (!arrow_schema) {
+        for (const auto& col : descriptor.schema.columns) {
+            if (detail::is_compound(col.data_type)) {
+                arrow_schema = detail::columns_to_arrow_schema(descriptor.schema.columns);
+                break;
+            }
+        }
+    }
+
+    if (arrow_schema) {
+        TableDescriptor arrow_desc = descriptor;
+        arrow_desc.schema.columns.clear();
+        auto ffi_desc = utils::to_ffi_table_descriptor(arrow_desc);
+        size_t schema_ptr = 0;
+        try {
+            schema_ptr = detail::export_arrow_schema(*arrow_schema);
+        } catch (const std::exception& e) {
+            return utils::make_client_error(e.what());
+        }
+        auto ffi_result =
+            admin_->create_table_arrow(ffi_path, ffi_desc, schema_ptr, ignore_if_exists);
+        return utils::from_ffi_result(ffi_result);
+    }
+
+    auto ffi_desc = utils::to_ffi_table_descriptor(descriptor);
     auto ffi_result = admin_->create_table(ffi_path, ffi_desc, ignore_if_exists);
     return utils::from_ffi_result(ffi_result);
 }
--- a/bindings/cpp/src/lib.rs
+++ b/bindings/cpp/src/lib.rs
@@ -17,17 +17,20 @@
 
 mod types;
 
+use std::collections::HashMap;
 use std::str::FromStr;
 use std::sync::{Arc, LazyLock, Mutex};
 use std::time::Duration;
 
+use arrow::ffi::{FFI_ArrowArray, FFI_ArrowSchema};
 use fluss as fcore;
 use fluss::PartitionId;
 use fluss::client::PrefixKeyLookuper;
 use fluss::error::Error;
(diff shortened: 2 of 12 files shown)


fresh-borzoni:

bindings/cpp/src/admin.cpp
```diff
@@ -58,8 +60,37 @@ Result Admin::CreateTable(const TablePath& table_path, const TableDescriptor& de
...
+    // A MAP/ROW column can't go through the flat FFI encoding, so the schema is
+    // sent over Arrow instead (explicit via FromArrow, or lowered from native
+    // columns). Rust derives the columns from it, so the flat columns are dropped
+    // here; primary keys / metadata still come from the descriptor.
+    std::shared_ptr<arrow::Schema> arrow_schema = descriptor.schema.arrow_schema;
+    if (!arrow_schema) {
+        for (const auto& col : descriptor.schema.columns) {
+            if (detail::is_compound(col.data_type)) {
+                arrow_schema = detail::columns_to_arrow_schema(descriptor.schema.columns);
+                break;
```
leekeiabstraction, line 73: Trying to understand why we break here, what if multiple columns are of row/map type?

leekeiabstraction (commented):
TY for the PR. Reviewing this in parts; will continue the rest tomorrow. Left a clarification question.

bindings/cpp/src/admin.cpp
fresh-borzoni: it's just a detection loop here, "is any datatype is complex/nested here in schema?", once the answer is -yes, we can short-circuit.

bindings/cpp/test/test_kv_table.cpp
```diff
@@ -212,24 +214,776 @@ TEST_F(KvTableTest, LookupWithNestedArrayArrayView) {
...
+        EXPECT_EQ(result2.GetInt32(0), 2);
+        for (size_t i = 1; i <= 9; ++i) {
+            EXPECT_TRUE(result2.IsNull(i)) << "column " << i << " should be null";
+        }
+    }
+
+    ASSERT_OK(adm.DropTable(table_path, false));
+}
+
+TEST_F(KvTableTest, LookupWithValueHandle) {
```
leekeiabstraction: Seems like a lot of the checks in LookupWithValueHandle and LookupComplexTypesMatrix are similar. Is it worth refactoring to minimise code to maintain?

bindings/cpp/test/test_log_table.cpp
leekeiabstraction, line 1795: Do we need MapWriterOverflowDetection test case as well?

bindings/cpp/test/test_kv_table.cpp
```diff
@@ -212,24 +214,776 @@ TEST_F(KvTableTest, LookupWithNestedArrayArrayView) {
...
+        {arrow::field("seq", arrow::int32()), arrow::field("label", arrow::utf8())});
+
+    auto arrow_schema = arrow::schema({
+        arrow::field("id", arrow::int32()),
+        arrow::field("m_str_int", arrow::map(arrow::utf8(), arrow::int32())),
+        arrow::field("m_str_row", arrow::map(arrow::utf8(), row_seq_label)),
+        arrow::field("m_str_map",
+                     arrow::map(arrow::utf8(), arrow::map(arrow::utf8(), arrow::int32()))),
+        arrow::field("m_str_arr", arrow::map(arrow::utf8(), arrow::list(arrow::int32()))),
+        arrow::field("arr_map", arrow::list(arrow::map(arrow::utf8(), arrow::int32()))),
```
leekeiabstraction, line 321: Should we have coverage for small and tiny ints?

leekeiabstraction (commented):
Thank you for the PR. Left some comments / questions.

bindings/cpp/test/test_kv_table.cpp
fresh-borzoni: Leftover, removed it and moved some checks in the matrix

bindings/cpp/test/test_log_table.cpp
fresh-borzoni: resolved

fresh-borzoni:
@leekeiabstraction Thank you for the review, addressed, PTAL 🙏

bindings/cpp/test/test_kv_table.cpp
fresh-borzoni: They are covered by all_datatypes checks, if we wish we can add checks, but then it's also needed to be handled in python/rust. I'll either skip or defer to a separate PR with parity across bindings.

fresh-borzoni: Hmm, actually it's C++ specific thing, so after another thought - it's worth adding still

bindings/cpp/src/lib.rs
```diff
@@ -17,17 +17,20 @@
 
 mod types;
 
+use std::collections::HashMap;
```
leekeiabstraction, line 20: Out of scope for this PR: I'm finding that this file has grown quite big and also some implementation verbose (e.g. could be simplified with macro) for effective human review. Do you think that this would benefit from being broken up?

leekeiabstraction (approved):
TY for the PR. LGTM.

bindings/cpp/src/lib.rs
fresh-borzoni: yes, we can split this per wrapper, good call 👍

PR #515, opened 2026-04-26
[client] Implement BatchScanner with limit-based scan

Arnav-panjla:
### Purpose

Fixes #316

Implement one-shot `BatchScanner` using `LimitScanRequest`.

---

### Brief change log

- Added `limit` to `TableScan`
- Added `create_batch_scanner`
- Implemented `BatchScanner` (eager RPC, single `poll_batch`)
- Added Arrow IPC (log) and KV → `RecordBatch` (PK) decoding
- Added projection support

---

### Tests

- All tests pass (`cargo test`)
- Verified with `cargo check`, `clippy`, `fmt`

---

### API and Format

**API Change:** Yes  
- New `BatchScanner`, `TableScan::limit`, `create_batch_scanner`  

**Storage Format:** No

---

### Documentation

No

PR #613, opened 2026-06-09
[python] Limit scan

fresh-borzoni:
closes https://github.com/apache/fluss-rust/issues/612

Python binding for the limit scan from #515. `table.new_scan().limit(n).create_bucket_batch_scanner(bucket)` returns a `BatchScanner` with `next_batch()`, `collect_all_batches()`, `to_arrow()` and `to_pandas()` -  a thin wrapper over the core `LimitBatchScanner`.

Porting projection turned up a core bug: `decode_log_batch` used projection pushdown, which assumes the server already projected, but limit scans don't project server side so it misread the buffers. Now decodes the full batch and projects after, like the KV path. Added the projection test #515 was missing.

### DIFF
--- a/bindings/python/src/table.rs
+++ b/bindings/python/src/table.rs
@@ -21,6 +21,7 @@ use arrow::array::RecordBatch as ArrowRecordBatch;
 use arrow::record_batch::RecordBatchReader as _;
 use arrow_pyarrow::{FromPyArrow, ToPyArrow};
 use arrow_schema::SchemaRef;
+use fcore::client::LimitBatchScanner;
 use fcore::metadata::{DataField, DataType, MapType, RowType};
 use fcore::row::binary_array::{FlussArray, FlussArrayWriter};
 use fcore::row::binary_map::{FlussMap, FlussMapWriter};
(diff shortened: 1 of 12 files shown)


fresh-borzoni:

bindings/python/src/table.rs
```diff
@@ -2780,6 +2842,116 @@ impl LogScanner {
...
+                .map_err(|e| FlussError::from_core_error(&e))?
+                .into_iter()
+                .map(|sb| Arc::new(sb.into_batch()))
+                .collect();
+            Python::attach(|py| LogScanner::batches_to_arrow_table(py, batches, &projected_schema))
+        })
+    }
+
+    /// Drain the scanner into a Pandas DataFrame.
+    fn to_pandas<'py>(&self, py: Python<'py>) -> PyResult<Bound<'py, PyAny>> {
```
charlesdong1991, line 2895: arrow table can call to_pandas directly to convert to a pandas dataframe, right? i wonder if it makes sense to use `to_arrow` here directly inside this function, and only do to_pandas in the end?

charlesdong1991 (approved):
Nice finding!! 👍 thanks!

bindings/python/src/table.rs
fresh-borzoni: Sure, we can 👍

PR #614, opened 2026-06-10
[c++]  Limit batch scanner

fresh-borzoni:
closes https://github.com/apache/fluss-rust/issues/615

C++ binding for the limit scan from #515. `table.NewScan().Limit(n).CreateBucketBatchScanner(bucket, scanner)` returns a `BatchScanner` with `NextBatch()`, `CollectAllBatches()` and `Bucket()`. It wraps the core `LimitBatchScanner` and hands the rows back as Arrow batches over the existing C Data Interface path, so there's little new plumbing.

### DIFF
--- a/crates/fluss/src/client/table/batch_scanner.rs
+++ b/crates/fluss/src/client/table/batch_scanner.rs
@@ -410,7 +410,7 @@ mod tests {
     use crate::row::binary::BinaryWriter;
     use crate::row::compacted::CompactedRowWriter;
     use crate::test_utils::build_table_info_with_columns;
-    use arrow::array::{Array, Int32Array, Int64Array};
+    use arrow::array::{Array, Int32Array, Int64Array, StringArray};
 
     fn build_two_col_table_info() -> TableInfo {
         build_table_info_with_columns(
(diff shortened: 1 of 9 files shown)


fresh-borzoni:

leekeiabstraction (approved):
LGTM, thank you for the PR!

crates/fluss/src/client/table/batch_scanner.rs
```diff
@@ -211,17 +202,21 @@ fn decode_log_batch(
...
+        concat_batches(&full_schema, batches.iter()).map_err(|e| Error::UnexpectedError {
             message: format!("Failed to concatenate log record batches: {e}"),
             source: None,
         })?
     };
 
-    Ok(take_last_rows(merged, base_offset, limit))
+    let (trimmed, base_offset) = take_last_rows(merged, base_offset, limit);
+    Ok((
+        project_batch(trimmed, table_info.get_row_type(), projected_fields)?,
```
charlesdong1991, line 217: this rust-core change now decodes full server respones and project client-side, which is right! 

Nit: can we add a test here using non-prefix projection which is never the case and not tested before? but completely nit, and can be follow-up pr or safely ignore, since it's just to avoid any regression since it is rust core change

charlesdong1991 (approved):
very nice! LGTM!! thanks again!

crates/fluss/src/client/table/batch_scanner.rs
fresh-borzoni: Sure, we have tests in bindings, but you are right, thank you

PR #617, opened 2026-06-14
[metrics] Add client metrics and prometheus example

charlesdong1991:
### Purpose
Document the implemented connection/writer/scanner client metrics and show how to collect them with a Prometheus exporter.

Linked issue: close #618

### DIFF
--- a/crates/examples/src/example_prometheus_metrics.rs
+++ b/crates/examples/src/example_prometheus_metrics.rs
@@ -1,0 +1,136 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+//! Exposes Fluss client metrics on a Prometheus scrape endpoint.
+//!
+//! Run a local cluster, then:
+//! ```shell
+//! cargo run -p fluss-examples --example example-prometheus-metrics
+//! curl http://localhost:9000/metrics
+//! ```
+//! The endpoint exposes `fluss_client_writer_*`, `fluss_client_scanner_*`, and
+//! `fluss_client_requests_*` series produced by the workload below.
+
+#[cfg(not(target_env = "msvc"))]
+#[global_allocator]
+static GLOBAL: tikv_jemallocator::Jemalloc = tikv_jemallocator::Jemalloc;
+
+use clap::Parser;
+use fluss::client::FlussConnection;
+use fluss::config::Config;
+use fluss::error::Result;
+use fluss::metadata::{DataTypes, Schema, TableDescriptor, TablePath};
+use fluss::row::{DataGetters, GenericRow};
+use metrics_exporter_prometheus::PrometheusBuilder;
+use std::time::Duration;
+
+#[tokio::main]
+pub async fn main() -> Result<()> {
+    // Install the global Prometheus recorder BEFORE creating any connection,
+    // writer, or scanner: the client caches metric handles on first use and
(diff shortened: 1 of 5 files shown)


charlesdong1991:
cc @fresh-borzoni @leekeiabstraction PTAL! 🙏

crates/examples/src/example_prometheus_metrics.rs
```diff
@@ -0,0 +1,136 @@
...
+        // can push it higher). This proves the recorder is wired up correctly.
+        if !verified {
+            let rendered = metrics_handle.render();
+            let sent = counter_value(&rendered, "fluss_client_writer_records_send_total");
+            assert!(
+                sent.is_some_and(|v| v >= rows_per_iter as f64),
+                "expected fluss_client_writer_records_send_total >= {rows_per_iter}, got {sent:?}\n{rendered}"
+            );
+            println!("self-check OK: records_send_total = {}", sent.unwrap());
+            verified = true;
```
leekeiabstraction, line 122: Do we need to break out of the loop once verification is done? As is, the example would run indefinitely (is that your intention?)

leekeiabstraction (commented):
TY for the PR. Left a question

crates/examples/src/example_prometheus_metrics.rs
charlesdong1991: Yes, it was the idea actually, because it's for prometheus demo, so how it usually works is the process should stays alive so people can scrape metric endpoint e.g. something like localhost:xxxx/metrics

if we break out of loop, then endpoint will be taken down, so users cannot scrape metrics anymore... 

this verify flag was just a check to confirm recorder is wired up, and not supposed to be used for stop condition. if users really want to stop, usually, they should ctrl-C which also reflects how such prometheus setup is working

i can add a bit more explanation in code to make it clearer if you think makes sense @leekeiabstraction 🙏

charlesdong1991: btw, maybe to avoid confusion, examples are not invoked in any CI, so this indefinite loop won't hang CI i believe

leekeiabstraction: I see, I may have misremembered that some examples are being ran during CI to ensure their correctness. This is fine if it is not ran during CI. If we do that eventually, it might be worth having it run for  something like 60s instead.

crates/examples/src/example_prometheus_metrics.rs
```diff
@@ -0,0 +1,136 @@
...
+            row.set_field(0, id);
+            row.set_field(1, "metrics demo");
+            append_writer.append(&row)?;
+            id += 1;
+        }
+        append_writer.flush().await?;
+
+        let scan_records = log_scanner.poll(Duration::from_secs(1)).await?;
+        let mut count = 0;
+        for record in scan_records {
```
leekeiabstraction: We are only verifying metrics for sent records, do we need the loop for polling records? If not, it might be worth removing this loop for a more concise example.

website/docs/user-guide/rust/metrics.md
```diff
@@ -0,0 +1,179 @@
...
+```
+
+A full, runnable program is available as the `example-prometheus-metrics` example
+in the `fluss-examples` crate.
+
+:::warning Install the recorder before writing or scanning
+The client caches metric handles the first time a writer or scanner is created,
+binding them to whichever recorder is installed at that moment. Install your global
+recorder **before** calling `FlussConnection::new` (ideally as the very first thing
+in `main`). If you install it after creating a writer or scanner, those metrics will
```
leekeiabstraction, line 50: This behaviour might catch user off guard. Is it worth making it so that an Error is returned instead?

leekeiabstraction (commented):
TY for clarifying. Left a couple more qs.

crates/examples/src/example_prometheus_metrics.rs
charlesdong1991: we need `poll` still because all scanner metrics are recorded inside poll(), but you are right that the per-field decode loop drives no metrics, so i replaced it!

website/docs/user-guide/rust/metrics.md
charlesdong1991: i think returning an error isn't workable because "no recorder installed" is a supported mode most users run in, so writer/scanner creation can't fail just because no recorder is present

Also metrics facade has no hook to detect "installed too late". The only alternative is to stop caching handles and re-resolve per record, and that will have hot-path cost so we should try to avoid.

wdyt? i can also some more explanation on that in this doc 🙏

leekeiabstraction (approved):
Ty for the PR and addressing comments/answering questions. LGTM!

PR #619, opened 2026-06-14
[metrics] Add client-side error and retry tracking metrics

charlesdong1991:
### Purpose
Add Rust client-side error/retry counters for production alerting: connections poisoned, metadata refreshes/errors, terminal writer errors (by error_kind), and scanner fetch errors (rpc vs bucket).

Java has error family on server, but not reachable from rust client.

Linked issue: close #620 

### Tests

All tests passed

### DIFF
--- a/crates/fluss/src/metrics.rs
+++ b/crates/fluss/src/metrics.rs
@@ -35,6 +35,11 @@ pub const LABEL_API_KEY: &str = "api_key";
 pub const LABEL_DATABASE: &str = "database";
 pub const LABEL_TABLE: &str = "table";
 
+/// Classifies an error counter sample. See the per-metric docs below for the
+/// value set each counter uses (writer terminal failures vs scanner fetch
+/// errors).
+pub const LABEL_ERROR_KIND: &str = "error_kind";
+
 // ---------------------------------------------------------------------------
 // Connection / RPC metrics
 //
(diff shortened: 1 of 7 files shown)


charlesdong1991:
@fresh-borzoni @leekeiabstraction PTAL 🙏

crates/fluss/src/metrics.rs
```diff
@@ -295,6 +363,27 @@ pub const WRITER_BYTES_SEND_TOTAL: &str = "fluss.client.writer.bytes_send.total"
...
+pub const WRITER_ERRORS_TOTAL: &str = "fluss.client.writer.errors.total";
+
+/// `error_kind` value for a non-retriable server error.
+pub(crate) const WRITER_ERROR_KIND_NON_RETRIABLE: &str = "non_retriable";
+/// `error_kind` value for a retriable error that exhausted its retries.
+pub(crate) const WRITER_ERROR_KIND_MAX_RETRIES_EXCEEDED: &str = "max_retries_exceeded";
+/// `error_kind` value for an idempotent retry abandoned after a writer-id reset.
+pub(crate) const WRITER_ERROR_KIND_WRITER_ID_CHANGED: &str = "writer_id_changed";
+/// `error_kind` value for a batch that failed to build locally (never sent).
+pub(crate) const WRITER_ERROR_KIND_LOCAL_BUILD: &str = "local_build";
```
leekeiabstraction, line 385: Do all of these metrics exist on Java client side? 

If not you may want to consider raising a short FIP (potentially including more new client side metrics that might be valuable) and discussion for the definitions and implementation of these metrics. Metrics are user facing, releasing and then removing these metrics at later release would be a breaking change so we'd want to be mindful about what we introduce. I think we would benefit from discussing and gathering feedback from the community.

leekeiabstraction (commented):
Thank you for the PR! Better observability makes sense to me but I've left a question.

crates/fluss/src/metrics.rs
charlesdong1991: > Do all of these metrics exist on Java client side?

No, they are not as i mentioned in the PR description, Java has things similar in server side, but rust client cannot get those... it is mainly inspired by those java side metrics, and i come up with some simple metrics for alerting accordingly.

> Metrics are user facing, releasing and then removing these metrics at later release would be a breaking change

You are right, i thought it's mainly on client side, so can be more flexible, but indeed better to share there to collect feedback first... I will think about how to draft discussion properly

leekeiabstraction: It's a good opportunity for looking at what other clients e.g. Kafka, kinesis etc. emit that might be useful for Fluss. I think it can be a relatively short but impactful FIP. We'll have to implement on Java side too though

fresh-borzoni (commented):
@charlesdong1991 Ty for the PR, it's a good point to think about this. It feels Java client never needed these metrics as it was mostly used by Flink, so errors were surfaced by task failure, then Flink metrics for this.

two things I think worth discussing related to current implementation:
  - the writer error_kind values mix a couple of axes and lean on internal terms (poisoned, writer_id_changed, local_build) - worth simplifying to operator-facing names reusable across clients as well.
  - scanner errors are per-table but writer errors are global/unlabeled, so it's a mismatch and it's really important to get the erroring table for the operator in general

Stepping back though - the bigger question is whether the Kafka metrics model itself is just dated. It's all flat counters: you get a number for errors, but never the why. Newer systems like Pulsar (and OpenTelemetry generally) have moved to tagging an operation with success/failed plus a short reason, so one metric gives you the breakdown and the error rate together. Maybe instead of porting more Kafka-style metrics, it's worth deciding if Fluss wants to move its client metrics to that more modern model as a whole, across both clients. 

That's a separate, bigger discussion than this PR, but this feels like a good moment to raise it for the bigger discussion, not only as part of metrics parity across clients.

charlesdong1991:
thanks for suggestions @fresh-borzoni @leekeiabstraction yeah, i am quite particularly interested in improving observability in general for Fluss as it is a big factor for adoption in production environment, i will seek for discussions in Fluss community first and help with that there first later next week once I am back from vacation.

Thanks a lot for suggestions, i really appreciate it!

PR #624, opened 2026-06-15
fix(test-cluster): manage containers via testcontainers' Docker client so start/stop share one runtime

beryllw:
### Purpose

Linked issue: close #623 

### Brief change log

- stop_cluster, FlussTestingCluster::stop, and the start pre-clean now force-remove via the shared client. A 404 (already gone) is treated as success; other errors warn instead of being silently swallowed.
- all_containers_exist lists containers through the client with an anchored name filter (now async).
- New run_blocking helper drives the async teardown from the synchronous stop() callers, working both inside a tokio runtime (async integration tests) and from the atexit cleanup handler.
- Imports the Docker types via the testcontainers::bollard re-export, so the crate takes no direct bollard dependency (Cargo.toml/Cargo.lock unchanged on balance).

### Tests

Verified against a live Podman cluster:
```
start            -> zookeeper / coordinator / tablet containers Up
start (again)    -> idempotent, no recreation, no name conflict
stop             -> all three containers removed
```

cargo build / cargo clippy clean for the crate (the 3 pre-existing main.rs readiness-poll lints are untouched and out of scope).

### DIFF
--- a/crates/fluss-test-cluster/src/lib.rs
+++ b/crates/fluss-test-cluster/src/lib.rs
@@ -18,12 +18,65 @@
 use fluss::client::FlussConnection;
 use fluss::config::Config;
 use std::collections::HashMap;
+use std::future::Future;
 use std::mem::ManuallyDrop;
 use std::sync::Arc;
 use std::time::Duration;
+use testcontainers::bollard::Docker;
+use testcontainers::bollard::query_parameters::{
+    ListContainersOptionsBuilder, RemoveContainerOptionsBuilder,
+};
 use testcontainers::core::ContainerPort;
+use testcontainers::core::client::docker_client_instance;
 use testcontainers::runners::AsyncRunner;
 use testcontainers::{ContainerAsync, GenericImage, ImageExt};
+
+/// testcontainers' own shared `bollard::Docker` — the same client (and daemon
+/// resolution) it uses for `image.start()`, so teardown and existence checks act on
+/// the exact daemon `start` created containers on. `None` (with a warning) if the
+/// client is unreachable.
+async fn docker_client() -> Option<Docker> {
+    match docker_client_instance().await {
+        Ok(docker) => Some(docker),
+        Err(e) => {
+            eprintln!("warning: cannot reach the testcontainers Docker client: {e}");
+            None
+        }
+    }
+}
+
+/// Force-removes a container by name. Best-effort: a 404 (already gone) is success;
+/// other errors warn rather than panic.
+async fn force_remove_container(docker: &Docker, name: &str) {
+    let options = RemoveContainerOptionsBuilder::default().force(true).build();
+    match docker.remove_container(name, Some(options)).await {
+        Ok(())
+        | Err(testcontainers::bollard::errors::Error::DockerResponseServerError {
+            status_code: 404,
+            ..
+        }) => {}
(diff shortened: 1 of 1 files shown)


crates/fluss-test-cluster/src/lib.rs
```diff
@@ -165,21 +218,32 @@ impl FlussTestingClusterBuilder {
...
+            filters.insert("name".to_string(), vec![format!("^{name}$")]);
+            let options = ListContainersOptionsBuilder::default()
+                .all(false)
+                .filters(&filters)
+                .build();
+            match docker.list_containers(Some(options)).await {
+                Ok(list) if !list.is_empty() => continue,
+                _ => return false,
+            }
+        }
```
leekeiabstraction, line 238: nit: Seems like we're making a list_containers per-container. Anyway to make this more efficient?

crates/fluss-test-cluster/src/lib.rs
```diff
@@ -493,22 +560,47 @@ impl FlussTestingCluster {
...
+        .filters(&filters)
+        .build();
+
+    let containers = match docker.list_containers(Some(options)).await {
+        Ok(containers) => containers,
+        Err(e) => {
+            eprintln!("warning: failed to list cluster containers: {e}");
+            return;
+        }
+    };
```
leekeiabstraction, line 596: nit: This looks similar to parts of `all_containers_exist`, can these be refactored?

leekeiabstraction (commented):
Thank you for the improvements. Left some small comments.

crates/fluss-test-cluster/src/lib.rs
```diff
@@ -493,22 +560,47 @@ impl FlussTestingCluster {
...
+    let Some(docker) = docker_client().await else {
+        return;
+    };
+
+    // Multiple values for the `name` filter are OR'd by the daemon; these prefixes
+    // cover zookeeper, coordinator, and any number of tablet servers.
+    let mut filters = HashMap::new();
+    filters.insert(
+        "name".to_string(),
+        vec![
```
charlesdong1991, line 579: not a concern, nor it will cause any issues here 

but just add a note: when using prefix here, if two different clusters share the prefix of `name` (e.g. one cluster has name of `test` and there other has `test2` or something), i think when stopping, it can accidentally stop both of them although purpose might be only stopping containers of `test` cluster

charlesdong1991 (approved):
Great improvement! 👍

leekeiabstraction (approved):
Approving to get this improvement merged. The earlier comments from me were nits. 

TY for your contribution!

crates/fluss-test-cluster/src/lib.rs
beryllw: Sorry for the late reply, +1.  I've already addressed it in a follow-up PR (#627).

crates/fluss-test-cluster/src/lib.rs
beryllw: I left it inline for now since the error handling differs slightly.

crates/fluss-test-cluster/src/lib.rs
beryllw: Thanks for review, good catch — fixed in a follow-up PR (https://github.com/apache/fluss-rust/pull/627). The prefix is only used for narrowing, and removal is guarded by an exact local name check, so test and test2 won't interfere.

PR #628, opened 2026-06-19
[rust] Sync Fluss 1.x proto definitions and ApiKey registry

gnuhpc:
## What

Syncs the Rust client's `flüss_api.proto` and `ApiKey` registry with Fluss 1.x, and updates pre-existing message wrappers that reference the renamed symbols so the crate still builds.

- Add all 1.x message types (ACLs, KV snapshots, producer offsets, cluster config, rebalance, server tags, etc.) and new optional fields (`rack`, `remote_data_dir`, `leader_epoch`, `agg_mode`, ...).
- Register 24 new `ApiKey` variants (1023–1064) in `api_key.rs`.
- Update `build.rs` prost-bytes config for the new proto fields.
- Add `None` defaults in `convert.rs` / `partition.rs`.
- Update pre-existing message wrappers that reference renamed proto fields / ApiKey variants so the build stays green:
  - `create_partition.rs`: `ignore_if_exists` -> `ignore_if_not_exists` (same wire field 3; matches Java's end-to-end value flow, no inversion)
  - `get_latest_lake_snapshot.rs`: `ApiKey::GetLatestLakeSnapshot` -> `GetLakeSnapshot` (same id 1032)
  - `list_databases.rs`: populate new `include_summary` field
  - `lookup.rs`: `PbLookupReqForBucket.key` -> `keys`; new `LookupRequest` fields

## Why

Foundation for the rest of the 1.x support stack. Each commit in this stacked PR series must leave `cargo build` green, so the proto/contract change ships together with the caller updates it requires.

## Notes

This is PR 1/6 of a stacked series (`pr/1` -> `pr/2` -> ... -> `pr/6`). The field renames are same-tag, so they are wire-compatible; the Java server's end-to-end handling was verified to preserve the original `ignore_if_exists` semantic.

### DIFF
--- a/crates/fluss/build.rs
+++ b/crates/fluss/build.rs
@@ -22,8 +22,9 @@ fn main() -> Result<()> {
     config.bytes([
         ".proto.PbProduceLogReqForBucket.records",
         ".proto.PbPutKvReqForBucket.records",
-        ".proto.PbLookupReqForBucket.key",
+        ".proto.PbLookupReqForBucket.keys",
         ".proto.PbPrefixLookupReqForBucket.keys",
+        ".proto.ScanKvResponse.records",
     ]);
     config.compile_protos(&["src/proto/fluss_api.proto"], &["src/proto"])?;
     Ok(())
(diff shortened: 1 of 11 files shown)


fresh-borzoni (approved):
@gnuhpc Thank you, LGTM

PR #629, opened 2026-06-19
[rust] Add RPC message wrappers for core DB/table operations

gnuhpc:
## What

Adds request/response message wrappers for the core 1.x DB/table admin RPCs: `AlterClusterConfigs`, `AlterDatabase`, `AlterTable`, `CreateAcls`, `DescribeClusterConfigs`, `DropAcls`, `GetTableStats`, `ListAcls`, `ListDatabaseSummaries`.

Each is a thin wrapper over the generated `proto::` types (from #628) wired into `rpc/message/mod.rs`. The admin client methods that consume them land in a later PR.

## Stack

Part **2/6** of the Fluss 1.x Rust support series, stacked on #628 (`pr/1-proto-sync`). Each branch in the stack compiles on top of its predecessor, so all PRs here target `main`; once #628 merges, this PR's diff automatically reduces to just the 9 message files unique to this branch.


crates/fluss/src/rpc/message/get_table_stats.rs
```diff
@@ -0,0 +1,48 @@
...
+    pub inner_request: proto::GetTableStatsRequest,
+}
+
+impl GetTableStatsRequest {
+    pub fn new(table_id: i64, buckets_req: Vec<proto::PbTableStatsReqForBucket>) -> Self {
+        GetTableStatsRequest {
+            inner_request: proto::GetTableStatsRequest {
+                table_id,
+                buckets_req,
+                target_columns: vec![],
```
fresh-borzoni: why do we hardcode target_columns: vec![]?
There's no way to ask for specific columns with this, is it intentional?

crates/fluss/src/rpc/message/alter_cluster_configs.rs
```diff
@@ -0,0 +1,44 @@
...
+// under the License.
+
+use crate::rpc::api_key::ApiKey;
+use crate::rpc::frame::{ReadError, WriteError};
+use crate::rpc::message::{ReadType, RequestBody, WriteType};
+use crate::{impl_read_type, impl_write_type, proto};
+use bytes::{Buf, BufMut};
+use prost::Message;
+
+#[derive(Debug, Default)]
```
fresh-borzoni: Can we drop #[derive(Default)] on the arg-taking requests? 

default() just builds empty(invalid requests), and the rest of the wrappers are #[derive(Debug)] only, so it's inconsistent

fresh-borzoni (changes requested):
@gnuhpc Thank you for the PR, we need to address a couple of issues here.

We shouldn't use `proto::*` in any public `FlussAdmin` signature. 
Please, define domain types mirroring Java's with conversion in the wrapper layer

crates/fluss/src/rpc/message/get_table_stats.rs
gnuhpc: Not intentional — fixed in 720d069. `GetTableStatsRequest::new` now takes `target_columns: Vec<i32>` and threads it through. Single existing call site (the integration test on #632) will pass an explicit `vec![]` to preserve current behavior. Same drop-silently-features pattern also fixed on #631 (`alter_table` was hardcoding `vec![]` for drop/rename/modify columns); will be in the next force-push on that PR.

crates/fluss/src/rpc/message/alter_cluster_configs.rs
gnuhpc: Done in 720d069. Dropped `#[derive(Default)]` on all 7 arg-taking wrappers added by this PR (`alter_cluster_configs`, `alter_database`, `alter_table`, `create_acls`, `drop_acls`, `get_table_stats`, `list_acls`).

Kept `Default` on the param-less ones (`describe_cluster_configs`, `list_database_summaries`) since they match the pre-existing `list_databases.rs` precedent and `default()` produces a valid request there — happy to drop those too if you'd prefer strict uniformity.

Same change applied to the 18 arg-taking wrappers on #630.

gnuhpc:
Addressed the `proto::*`-in-public-API feedback in 720d069. Summary of what changed:

**Domain types introduced** under `crates/fluss/src/metadata/` (mirroring Java with `to_pb` / `from_pb` conversion in the wrapper layer):

- `config.rs` — `AlterConfig` + `AlterConfigOpType` (Set/Delete/Append/Subtract)
- `acl.rs` — `AclInfo` / `AclFilter` + `ResourceType` (Any/Cluster/Database/Table), `OperationType` (Any/All/Read/Write/Create/Drop/Alter/Describe), `PermissionType` (Any/Allow)
- `table_change.rs` — `AddColumn` / `DropColumn` / `RenameColumn` / `ModifyColumn` + `ColumnPositionType` (currently `Last` only, matching Java's `ColumnPositionType.java` — proto reserves `First=1`, `After=3` for future use)
- `table_stats.rs` — `BucketStatsRequest`

Enum int codes are taken verbatim from the Java source (`AlterConfigOpType.java`, `ResourceType.java`, `OperationType.java`, `PermissionType.java`, `ColumnPositionType.java`).

**Wrapper changes**: all 7 arg-taking wrappers' `new()` now accept domain types and call `.to_pb()` to populate `inner_request`. Wrappers still hold `proto::*` internally (matches the existing `PartitionSpec`/`DatabaseDescriptor` pattern).

**Stack-wide sweep**: since the same `proto::*`-leak class of issue exists on #630 and #631 (the FlussAdmin admin client surface in #631 was where most leaks lived), I'm restacking the rest of the chain with the same cleanup now — will post here once each PR is force-pushed.

**Tests**: added `*_pb_roundtrip` unit tests for each new domain type (matching the `PartitionSpec`/`DatabaseDescriptor` test style). `cargo test -p fluss-rs --lib` passes 544 (15 new).

Thanks for the review!

fresh-borzoni (approved):
@gnuhpc Thank you for the changes, LGTM
I pushed commit to tighten visibility for protos in your new code and a couple of pre-existing issues. Hope it's okay with you

PR #635, opened 2026-06-22
[rust] Apply TableId/PartitionId/BucketId type aliases to write path

gnuhpc:
## Summary

- Replace raw `i64`/`i32` with `TableId`/`PartitionId`/`BucketId` type aliases in the write path (`sender.rs`, `accumulator.rs`, `bucket_assigner.rs`) and RPC messages (`produce_log.rs`, `put_kv.rs`)
- Fix `cluster.rs` getter return types to use `TableId` instead of raw `i64`
- Align with the convention already used in newer modules (`lookup.rs`, `list_offsets.rs`, `prefix_lookup.rs`)

No behavioral change — these are type aliases (`type TableId = i64`, etc.), so this is purely a readability and consistency improvement.

## Test plan

- [x] `cargo build` — full workspace compiles
- [x] `cargo clippy --all-targets -- -D warnings` — zero warnings
- [x] `cargo fmt --all -- --check` — clean
- [x] `cargo test --workspace` — all 545 tests pass (existing tests cover the modified paths)

### DIFF
--- a/crates/fluss/src/client/table/remote_log.rs
+++ b/crates/fluss/src/client/table/remote_log.rs
@@ -988,10 +988,10 @@ mod tests {
     use super::*;
     use crate::metadata::TablePath;
     use crate::test_utils::test_scanner_metrics;
+    use crate::{BucketId, TableId};
     use std::sync::atomic::{AtomicUsize, Ordering};
 
-    /// Helper function to create a TableBucket for testing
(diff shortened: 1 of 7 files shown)


fresh-borzoni (approved):
@gnuhpc LGTM, thank you for the PR

PR #638, opened 2026-06-24
[cpp] Encode CreateTable column types losslessly

fresh-borzoni:
closes https://github.com/apache/fluss-rust/issues/637

Now the column types go to the server natively and are stored exactly as declared.

### DIFF
--- a/bindings/cpp/src/types.rs
+++ b/bindings/cpp/src/types.rs
@@ -19,6 +19,9 @@ use crate::ffi;
 use anyhow::{Result, anyhow};
 use arrow::array::Array;
 use arrow::ffi::{FFI_ArrowArray, FFI_ArrowSchema};
+use fcore::metadata::{
+    ArrayType, Column, DataField, DataType, DataTypes, DecimalType, MapType, RowType,
+};
 use fcore::row::Datum;
 use fluss as fcore;
 use std::borrow::Cow;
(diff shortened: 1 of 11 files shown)


fresh-borzoni:
@charlesdong1991 @leekeiabstraction PTAL, pls

fresh-borzoni:
@loserwang1024 @naivedogger Since you expressed an interest in C++ bindings PTAL :)

website/docs/user-guide/cpp/data-types.md
```diff
@@ -85,12 +85,9 @@ auto schema = fluss::Schema::NewBuilder()
...
 ```
 
-:::note Arrow escape hatch
-If you already have an `arrow::Schema`, pass it directly with
-`fluss::Schema::FromArrow(arrow_schema, /*primary_keys=*/{"id"})`. It's
-equivalent — the native factories above lower to the same Arrow types
-internally, without pulling Arrow into your code.
-:::
+Column types are sent to the server exactly as declared: precision, scale,
+length, per-level nullability, and `ROW` field names all round-trip losslessly
```
charlesdong1991, line 89: does it hold true for Time? i think precision will change still for time?

bindings/cpp/src/types.rs
```diff
@@ -44,229 +47,184 @@ pub const DATA_TYPE_ARRAY: i32 = 17;
...
+        DATA_TYPE_ROW => {
+            let mut fields = Vec::with_capacity(node.child_count as usize);
+            for _ in 0..node.child_count {
+                let field_name = nodes
+                    .get(*cursor)
+                    .ok_or_else(|| anyhow!("ROW field missing from type tree"))?
+                    .field_name
+                    .clone();
+                let field_type = nodes_to_data_type(nodes, cursor)?;
+                fields.push(DataField::new(field_name, field_type, None));
```
charlesdong1991, line 98: so we will drop description? i guess it is intended?

bindings/cpp/test/test_ffi_converter.cpp
```diff
@@ -18,223 +18,220 @@
...
+using fluss::Column;
+using fluss::DataType;
+using fluss::TypeId;
 
-TEST(FfiConverterTest, RejectsArrayWithoutElementType) {
-    auto col = MakeArrayColumn(1, 0);
-    EXPECT_THROW((void)fluss::utils::from_ffi_column(col), std::runtime_error);
+// Encode a column to the FFI node arena and decode it back, exercising the
+// full create-table / get-table-info type transport.
+Column RoundTrip(const Column& col) {
```
charlesdong1991, line 34: Nit: since the targets are precision, comment and char, can we add an integration test that create a table with those type, and then assert get-table-info returns exactly those values?

charlesdong1991 (commented):
very nice one! just some nit comments

fresh-borzoni:
move to fluss main repo

PR #643, opened 2026-06-27
[tests] Add DynamicWriteBatchSizeEstimator integration tests

pbanakar:
### Purpose

Linked issue: close #539

Adds integration tests for `DynamicWriteBatchSizeEstimator` to verify end-to-end write correctness through a real cluster for all four scenarios described in the issue.

### Brief change log

- Added `crates/fluss/tests/integration/dynamic_batch_size.rs` with 4 integration tests covering:
  - Many small rows cause the estimator to shrink batch size toward min
  - Rows close to batch size cause the estimator to grow toward max
  - Disabled config keeps the static `writer_batch_size` unchanged
  - Concurrent writers from separate connections don't corrupt estimator state
- Registered the new module in `crates/fluss/tests/test_fluss.rs`

### Tests

- `small_rows_shrink_batch_size` — writes 200 tiny rows with dynamic sizing enabled; verifies all writes succeed
- `large_rows_grow_batch_size` — writes rows filling >80% of batch capacity; verifies all writes succeed
- `disabled_keeps_static_batch_size` — writes with  `writer_dynamic_batch_size_enabled = false`; verifies all writes succeed
- `concurrent_writers_dont_corrupt_state` — spawns 4 concurrent writer tasks each with its own connection; verifies all writes succeed without errors

### API and Format

No changes to API or storage format.

### Documentation

No new feature introduced. This PR only adds integration tests for an existing feature.

### DIFF
--- a/crates/fluss/tests/integration/dynamic_batch_size.rs
+++ b/crates/fluss/tests/integration/dynamic_batch_size.rs
@@ -1,0 +1,241 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#[cfg(test)]
+mod dynamic_batch_size_test {
+    use crate::integration::utils::{create_table, get_shared_cluster, wait_for_table_ready};
+    use fluss::client::FlussConnection;
+    use fluss::config::Config;
+    use fluss::metadata::{DataTypes, Schema, TableDescriptor, TablePath};
+    use fluss::row::{Datum, GenericRow};
+
+    fn make_config(
+        bootstrap_servers: &str,
+        batch_size: i32,
+        dynamic_enabled: bool,
+        dynamic_min: i32,
+    ) -> Config {
+        Config {
+            bootstrap_servers: bootstrap_servers.to_string(),
+            writer_acks: "all".to_string(),
+            writer_batch_size: batch_size,
+            writer_dynamic_batch_size_enabled: dynamic_enabled,
+            writer_dynamic_batch_size_min: dynamic_min,
+            ..Config::default()
+        }
+    }
+
+    fn log_table_descriptor() -> TableDescriptor {
+        TableDescriptor::builder()
+            .schema(
+                Schema::builder()
+                    .column("id", DataTypes::int())
+                    .column("payload", DataTypes::string())
+                    .build()
+                    .expect("Failed to build schema"),
+            )
+            .build()
+            .expect("Failed to build table descriptor")
+    }
+
+    fn make_row(id: i32, payload: &str) -> GenericRow<'static> {
+        let mut row = GenericRow::new(2);
+        row.set_field(0, Datum::Int32(id));
+        row.set_field(1, Datum::String(payload.to_string().into()));
+        row
+    }
+
+    /// Many small rows should cause the estimator to shrink the batch size toward min.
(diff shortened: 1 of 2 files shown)


Prajwal-banakar:
Hi @charlesdong1991 @fresh-borzoni could you please help review this!? The failing Elixir check appears to be unrelated to this PR's changes, PTAL.

crates/fluss/tests/integration/dynamic_batch_size.rs
```diff
@@ -0,0 +1,241 @@
...
+            .create_writer()
+            .expect("Failed to create writer");
+
+        // Write many tiny rows — each well below 50% of the batch size, so
+        // the estimator should shrink the target toward min after each drain.
+        for i in 0..200 {
+            let row = make_row(i, "x");
+            writer.append(&row).expect("Failed to append row");
+        }
+        writer.flush().await.expect("Failed to flush");
```
charlesdong1991, line 137: i doubt this test will do the job (based on the comment and test func name) 🤔 since it only asserts flush succeeds, but don't observe estimated batch size, if a regression happens which disables shrinking, that will still pass i guess

so effectively, it is more like checking if dynamic write craches, but not end-to-end test that checks shrinking batch size towards min IMHO

crates/fluss/tests/integration/dynamic_batch_size.rs
```diff
@@ -0,0 +1,241 @@
...
+            .expect("Failed to create writer");
+
+        // Write rows with a payload that fills well above 80% of the batch.
+        // Repeat to give the estimator multiple drain cycles to grow.
+        let large_payload = "A".repeat(220 * 1024); // ~220 KB per row
+        for i in 0..5 {
+            let row = make_row(i, &large_payload);
+            writer.append(&row).expect("Failed to append row");
+        }
+        writer.flush().await.expect("Failed to flush");
```
charlesdong1991: similar here, i don't think in this case (starting with above 80% of max) will observe growth, i think we should rework it to shrink-then-grow with a read back assertion for an actual end-to-end test

crates/fluss/tests/integration/dynamic_batch_size.rs
```diff
@@ -0,0 +1,241 @@
...
+            let row = make_row(i, "static");
+            writer.append(&row).expect("Failed to append row");
+        }
+        writer.flush().await.expect("Failed to flush");
+    }
+
+    /// Multiple concurrent writers to the same table should not corrupt the estimator.
+    /// Each writer uses its own connection; all writes must succeed.
+    #[tokio::test]
+    async fn concurrent_writers_dont_corrupt_state() {
```
charlesdong1991: as mentioned in comment and in spawning below, these will be 4 independent connections/estimators, so there will be no shared state to corrupt, so i wonder if this func name is intended or not? Or what actually do we intend to test?

charlesdong1991 (changes requested):
Thanks for the PR! Left some comments, PTAL

crates/fluss/tests/integration/dynamic_batch_size.rs
pbanakar: Added read_back_records() helper that polls bucket 0 from EARLIEST_OFFSET and collects all records — every test now asserts the exact record count, so a silent regression would cause data loss and fail the assertion.

crates/fluss/tests/integration/dynamic_batch_size.rs
pbanakar: Renamed large_rows_grow_batch_size to shrink_then_grow_batch_size — Phase 1 drives the estimator down with 100 tiny rows, Phase 2 sends rows filling ~200 KB (well above 80% of the shrunk target) to trigger growth, both phases are read back.

crates/fluss/tests/integration/dynamic_batch_size.rs
pbanakar: Fixed concurrent_writers_dont_corrupt_state to concurrent_appends_share_estimator_without_corruption — all 4 tasks now share a single Arc<AppendWriter> from one connection, so they share the same RecordAccumulator and estimator, and concurrent access to shared state is actually exercised.

Prajwal-banakar:
Hi @charlesdong1991  addressed all three comments, PTAL another look when you have some time, thanks!

fresh-borzoni (commented):
@Prajwal-banakar Nice work, but I think Charles's first point still stands thougha as  asserting the record count only proves no data was lost, it doesn't observe the batch size. 

If shrinking regressed, all rows would still round-trip and the test would pass. Simplest fix: expose the estimator's current size via a pub(crate) accessor and assert it moved toward min/max.

fresh-borzoni:
moved to main repo and merged there

PR #645, opened 2026-06-28
fix: validate advertised API version ranges

Herrtian:
﻿## Summary

- Validate `ApiVersionsResponse` `api_key`, `min_version`, and `max_version` values before narrowing them to `i16`.
- Return `Error::UnsupportedVersion` with the offending field name and value instead of panicking on malformed server metadata.
- Propagate validation failures from the API versions handshake.

## Validation

- `cargo fmt --check`
- `git diff --check`
- `cargo check -p fluss-rs`
- `cargo test -p fluss-rs server_api_versions --lib`

Closes #535


Herrtian:
Closing this draft.

PR #646, opened 2026-06-29
[elixir] ci: bump OTP to 28.5.0.2

nicolazar:
### Purpose

This PR fixes the CI for Elixir bindings.

### Brief change log

The CI fails with:

```
Could not start Hex. Try fetching a new version with "mix local.hex" or uninstalling it with "mix archive.uninstall hex.ez"
** (MatchError) no match of right hand side value
{:failed_to_start_child, Hex.State,
         {:undef,
          [
            {:re, :import,
             [
               {:re_exported_pattern,
               ...
```

The problem here is essentially a toolchain version mismatch: the latest version of `hex` - 2.5.0 (picked by the CI action) had added a new feature that uses a regex for a duration parser, but compiled under Elixir 1.19 on OTP 28, that regex is serialized in OTP 28's new `export` form and reconstructed at load time via `re:import/1`. However, this function only since OTP 28.1 (hence the `undef`). 

So bumping the OTP version (latest 28.x) should fix this issue.

### DIFF
--- a/.github/workflows/build_and_test_elixir.yml
+++ b/.github/workflows/build_and_test_elixir.yml
@@ -43,7 +43,7 @@ jobs:
     timeout-minutes: 60
     runs-on: ubuntu-latest
     env:
-      OTP_VERSION: "28.0.2"
+      OTP_VERSION: "28.5.0.2"
       ELIXIR_VERSION: "1.19.5"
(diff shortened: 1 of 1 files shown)


fresh-borzoni (approved):
@nicolazar Thank you for the PR, makes sense 👍

PR #654, opened 2026-07-02
[rust] Support schema evolution for log scanner.

loserwang1024:
### Purpose

Linked issue: close https://github.com/apache/fluss-rust/issues/547

### DIFF
--- a/crates/fluss/src/client/table/log_fetch_buffer.rs
+++ b/crates/fluss/src/client/table/log_fetch_buffer.rs
@@ -18,6 +18,7 @@
 use arrow::array::RecordBatch;
 use parking_lot::Mutex;
 
+use crate::client::table::read_context_resolver::ReadContextResolver;
 use crate::client::table::remote_log::{
     PrefetchPermit, RemoteLogDownloadFuture, RemoteLogFile, RemoteLogSegment,
 };
@@ -85,7 +86,7 @@ pub trait PendingFetch: Send + Sync {
 
 /// Thread-safe buffer for completed fetches
 pub struct LogFetchBuffer {
-    read_context: ReadContext,
+    resolver: Arc<ReadContextResolver>,
     completed_fetches: Mutex<VecDeque<Box<dyn CompletedFetch>>>,
     pending_fetches: Mutex<HashMap<TableBucket, VecDeque<Box<dyn PendingFetch>>>>,
     next_in_line_fetch: Mutex<Option<Box<dyn CompletedFetch>>>,
@@ -94,9 +95,9 @@ pub struct LogFetchBuffer {
 }
 
 impl LogFetchBuffer {
-    pub fn new(read_context: ReadContext) -> Self {
+    pub fn new(resolver: Arc<ReadContextResolver>) -> Self {
         Self {
-            read_context,
+            resolver,
             completed_fetches: Mutex::new(VecDeque::new()),
             pending_fetches: Mutex::new(HashMap::new()),
             next_in_line_fetch: Mutex::new(None),
@@ -168,7 +169,7 @@ impl LogFetchBuffer {
             api_error,
             fetch_error_context,
             fetch_offset,
-            self.read_context.clone(),
+            Arc::clone(&self.resolver),
         );
         self.completed_fetches
             .lock()
@@ -226,7 +227,7 @@ impl LogFetchBuffer {
                 table_bucket.clone(),
                 error,
                 -1,
-                self.read_context.clone(),
+                Arc::clone(&self.resolver),
             );
             completed_to_push.push(Box::new(error_fetch));
         }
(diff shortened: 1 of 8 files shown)


loserwang1024:
@fresh-borzoni , CC

crates/fluss/src/client/table/read_context_resolver.rs
```diff
@@ -0,0 +1,186 @@
...
+        let effective_id = if self.projected_fields.is_some() {
+            self.initial_schema_id
+        } else {
+            schema_id
+        };
+
+        let guard = self.contexts.read();
+        guard.get(&effective_id).map(|ctx| {
+            if is_remote {
+                ctx.remote.clone()
```
fresh-borzoni: this returns an owned ReadContext clone per batch, where it used to be a borrow. Fine in the no-projection case, but with projection it deep-copies the Vecs on every batch. 
Could we store Arc<ReadContext> and return an Arc clone? Then it's a single refcount bump and the projection Vecs aren't copied this way

crates/fluss/src/client/table/scanner.rs
```diff
@@ -1581,7 +1607,7 @@ impl LogFetcher {
                     Self::pending_remote_fetches(
                         remote_log_downloader.clone(),
                         log_fetch_buffer.clone(),
-                        remote_read_context.clone(),
+                        Arc::clone(&resolver),
```
fresh-borzoni, line 1634: Schema prewarm only runs on the local path, not here. So reading tiered data written under an older schema fails with No ReadContext found for schema_id N.

The local branch below fetches missing schemas before decoding and the remote path needs the same.
WDYT?

crates/fluss/src/client/table/log_fetch_buffer.rs
```diff
@@ -518,7 +525,8 @@ impl DefaultCompletedFetch {
             };
 
             let log_batch = log_batch_result?;
-            let mut record_batch = log_batch.record_batch(&self.read_context)?;
+            let read_context = self.resolve_context_for_batch(&log_batch)?;
```
fresh-borzoni: After a schema change, batches here have different column counts, but the reader advertises one fixed schema. Consumers like DataFusion/pyarrow crash on that.
Should we pad each batch up to the advertised schema with null columns for the missing ones. 
Also probably needs a test, the IT only covers the row scanner

fresh-borzoni (commented):
@loserwang1024 Thank you for the PR, left comments, PTAL

loserwang1024:
@fresh-borzoni 


 Thank you very much for the thorough review. You were absolutely right that schema prewarming only covered the local path, while tiered/remote data could still fail with `No ReadContext found for schema_id N`.

 I explored several approaches because the remote path behaves differently from the local one. Local records are included directly in the fetch response, so their schema IDs can be collected and prewarmed before decoding. For tiered data, however, the fetch response only contains remote segment metadata. The actual file must first be downloaded and is then read lazily, batch by batch. Discovering all schema IDs in advance would require scanning the entire downloaded segment before returning any data, which would defeat the file-backed streaming behavior and increase both latency and memory usage.

 My first attempt in commit `04eb30dc` therefore fetched the schema lazily when the concrete batch was being decoded. However, schema resolution was still invoked from a synchronous decoding path. It used a Tokio runtime handle from another thread and then synchronously joined that thread. On a current-thread Tokio runtime, this could deadlock because the blocked runtime thread was also required to drive the schema RPC.

 The latest fix keeps the per-batch lazy-resolution approach but moves the asynchronous work to the scanner layer:

 - The synchronous decoder returns `SchemaRequired(schema_id)` when it encounters an uncached schema.
 - The current raw batch is retained instead of being discarded or advancing through the remote file.
 - The asynchronous scanner fetches and registers the schema without blocking the Tokio runtime.
 - The same batch is then retried and decoded.

 This allows both local and remote paths to resolve schemas on demand without preloading the entire remote segment and without blocking the runtime. I also added coverage for both dynamic-schema and fixed-schema modes.

 Thanks again for catching this and for the very detailed review.

fresh-borzoni (approved):
@loserwang1024 Nice changes 👍 

one note:
to_record_batch_reader() gives back an Arrow RecordBatchReader that has to expose one fixed schema. So with the default, a schema change mid-scan produces batches of different widths and DataFusion/pyarrow crash on it.
Since that reader is bounded anyway, conforming to the start schema is the same thing DataFusion/Iceberg/Delta do, you just re-run to pick up new columns. 

So should we default the batch scanner to fixed and leave the row scanner as-written? with_fixed_schema can still override either.

And let's move this to fluss main repo now as well

fresh-borzoni:
@loserwang1024 I'm a bit confused, shouldn't this be moved to https://github.com/apache/fluss since we merged the repos?
🤔

